Issue #13, opened 2016-04-04
Upgrade Scanamo


cb372:
So we can use the fancy new query API

PR #18, opened 2016-05-02
Upgrade Scanamo to 0.4.0

philwills:
There's a lot more hygiene that could be done, but this seems to work and I'd rather not have to deal with conflicts against *all* the files.

### DIFF
--- a/app/controllers/Amigo.scala
+++ b/app/controllers/Amigo.scala
@@ -3,21 +3,26 @@ package controllers
 import akka.stream.scaladsl.Source
 import com.gu.googleauth.GoogleAuthConfig
 import _root_.packer.{ PackerConfig, PackerRunner }
+import com.gu.scanamo.ops.{ ScanamoOps, ScanamoOpsA }
+import com.gu.scanamo.{ Scanamo, ScanamoAsync }
 import models._
 import data._
 import play.api.data.Form
 import play.api.data.Forms._
 import play.api.i18n.{ I18nSupport, MessagesApi }
 import play.api.libs.EventSource
-import play.api.libs.iteratee.{ Enumerator, Concurrent }
+import play.api.libs.iteratee.{ Concurrent, Enumerator }
 import event._
-
 import play.api.mvc._
 
 class Amigo(
     eventsSource: Source[BakeEvent, _],
     val authConfig: GoogleAuthConfig,
-    val messagesApi: MessagesApi)(implicit dynamo: Dynamo, packerConfig: PackerConfig, eventBus: EventBus) extends Controller with AuthActions with I18nSupport {
+    val messagesApi: MessagesApi,
+    bakes: Bakes,
+    recipes: Recipes,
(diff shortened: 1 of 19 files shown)


philwills:
Fixes #13

app/controllers/Amigo.scala
```diff
@@ -168,6 +175,8 @@ class Amigo(
       .via(EventSource.flow)
     Ok.chunked(source).as("text/event-stream")
   }
+
+  def exec[T](ops: ScanamoOps[T]) = Scanamo.exec(dynamo.client)(ops)
```
cb372: Make this private? Also a type annotation would be nice, as I have no idea what it returns.

app/data/BakeLogs.scala
```diff
@@ -19,17 +25,7 @@ object BakeLogs {
...
-    val queryRequest = new QueryRequest(tableName)
-      .withKeyConditionExpression("#bakeId = :bakeId")
-      .withExpressionAttributeNames(Map("#bakeId" -> "bakeId").asJava)
-      .withExpressionAttributeValues(Map(":bakeId" -> BakeId.dynamoFormat.write(bakeId)).asJava)
-      .withScanIndexForward(true) // return oldest logs first
-    val items: Iterable[Option[ValidatedNel[DynamoReadError, BakeLog]]] =
-      dynamo.client.query(queryRequest).getItems.asScala.map { item => Scanamo.from[BakeLog](new GetItemResult().withItem(item)) }
-    items.flatMap(_.flatMap(_.toOption))
+  def list(bakeId: BakeId): ScanamoOps[Iterable[BakeLog]] = {
+    table.query('bakeId -> bakeId).map(_.flatMap(_.toOption))
```
cb372: Much better!

So Scanamo queries with `scanIndexforward = true` by default?

app/data/Bakes.scala
```diff
@@ -26,40 +32,27 @@ object Bakes {
...
-      r <- recipe.toIterable
-      dbModel <- dbModels.flatMap(_.flatMap(_.toOption))
+      r <- recipes.findById(recipeId)
+      dbModels <- table.query(('recipeId -> recipeId).descending)
     } yield {
-      Bake.db2domain(dbModel, r)
+      for {
+        recipe <- r.toIterable
+        dbModel <- dbModels.flatMap(_.toOption).take(limit)
+      } yield Bake.db2domain(dbModel, recipe)
```
cb372: A for inside a yield looks a bit weird. Could make the second for-comp a local def?

cb372:
Thanks, I've been waiting for you to do this for me 😁 

It's a shame you had to change all the DAO objects (`Bakes`, etc.) into classes. I quite liked having them as objects. How about storing the Scanamo `Table` instances in `Dynamo.Tables` and letting the DAO objects look them up implicitly, like they were doing for the table names.

Apart from that, looks fine to me.

PR #21, opened 2016-05-16
Rv xenial for grid

Reettaphant:


app/packer/PackerBuildConfigGenerator.scala
```diff
@@ -46,8 +46,8 @@ object PackerBuildConfigGenerator {
     val provisioners = Seq(
       // bootstrap Ansible
       PackerProvisionerConfig.executeRemoteCommands(Seq(
+        "sleep 10",
```
cb372: It turns out that other people have already encountered this problem. See https://github.com/mitchellh/packer/issues/2639

Let's replace the naive sleep with the following command:

```
"while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done"
```

And please add a comment linking to the issue.

cb372:
Looks good apart from that 👍 

Thanks for being an early adopter!

PR #30, opened 2016-05-20
Implement scheduled bakes

### LINKED ISSUE #4: Scheduled bakes
For each recipe, you should have the option of specifying "bake this every N days"

cb372:
* Add a scheduler based on Quartz
* Add the corresponding form field to the UI

I decided to just let users input a raw cron expression in the UI. We're all adults here.

Fixes #4


app/components/AppComponents.scala
```diff
@@ -78,10 +79,19 @@ class AppComponents(context: Context)
 
   val prism = new Prism(wsClient)
 
+  val scheduledBakeRunner = {
+    val enabled = identity.stage == "PROD" // don't run scheduled bakes on dev machines
```
adamnfish: Is the compile-time-DI wiring class the place for logic like this?

app/schedule/BakeScheduler.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.quartz.impl.StdSchedulerFactory
+import org.quartz.{ TriggerKey, JobKey, JobDataMap }
+import play.api.Logger
+
+class BakeScheduler(scheduledBakeRunner: ScheduledBakeRunner) {
+
+  private val scheduler = StdSchedulerFactory.getDefaultScheduler()
+
+  def initialise(recipes: Iterable[Recipe]): Unit = {
+    recipes.flatMap(r => r.bakeSchedule.map(s => (r.id, s))).foreach {
```
adamnfish: It seems like one of the `map` and `foreach` is redundant here - all the map is doing is putting the data that's already in scope into a tuple that you then immediately extract. Would this be clearer as just a nested foreach?

```scala
recipes.foreach { recipe =>
  recipe.bakeSchedule.foreach { schedule =>
    scheduleBake(recipe.id, schedule)
  }
}
```

philwills:
Dammit, I'd just got the Scanamo upgrade PR rebased against master. 😿

cb372:
You snooze, you lose!

PR #32, opened 2016-05-23
Scanamo 0.5.0

philwills:
The `exec`s littered everywhere are horrid, but I don't want to leave this and have more merge 'fun'.

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -5,13 +5,14 @@ import akka.typed._
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.auth.{ InstanceProfileCredentialsProvider, AWSCredentialsProviderChain }
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
+import com.amazonaws.services.dynamodbv2.{ AmazonDynamoDB, AmazonDynamoDBClient }
 import com.gu.cm.{ ConfigurationLoader, Identity }
 import com.gu.googleauth.GoogleAuthConfig
+import data._
 import org.joda.time.Duration
 import play.api.ApplicationLoader.Context
 import play.api.libs.streams.Streams
-import play.api.{ Configuration, BuiltInComponentsFromContext }
+import play.api.{ BuiltInComponentsFromContext, Configuration }
 import play.api.i18n.I18nComponents
 import play.api.libs.iteratee.Concurrent
 import play.api.libs.ws.ahc.AhcWSComponents
@@ -44,21 +45,27 @@ class AppComponents(context: Context)
       new ProfileCredentialsProvider(),
       new InstanceProfileCredentialsProvider
     )
-    val region = Regions.fromName(configuration.getString("aws.region").getOrElse("eu-west-1"))
-    val dynamoClient: AmazonDynamoDBClient = new AmazonDynamoDBClient(awsCreds).withRegion(region)
-    new Dynamo(dynamoClient, identity.stage)
+    val region = configuration.getString("aws.region").map(Regions.fromName).getOrElse(Regions.EU_WEST_1)
+    val dynamoClient: AmazonDynamoDB = new AmazonDynamoDBClient(awsCreds).withRegion(region)
+    new Dynamo(dynamoClient, identity)
   }
   dynamo.initTables()
 
+  val baseImages = new BaseImages(identity)
+  val recipes = new Recipes(identity, baseImages)
+  val bakes = new Bakes(identity, recipes)
+  val bakeLogs = new BakeLogs(identity)
+
   val (eventsEnumerator, eventsChannel) = Concurrent.broadcast[BakeEvent]
   val eventsSource = Source.fromPublisher(Streams.enumeratorToPublisher(eventsEnumerator))
+  val behaviours = new Behaviours(bakes, bakeLogs)
   val eventBusActorSystem = {
     val eventListeners = Map(
-      "channelSender" -> Props(Behaviours.sendToChannel(eventsChannel)),
-      "logWriter" -> Props(Behaviours.writeToLog),
-      "dynamoWriter" -> Props(Behaviours.writeToDynamo)
+      "channelSender" -> Props(behaviours.sendToChannel(eventsChannel)),
+      "logWriter" -> Props(behaviours.writeToLog),
+      "dynamoWriter" -> Props(behaviours.writeToDynamo)
     )
-    ActorSystem[BakeEvent]("EventBus", Props(Behaviours.guardian(eventListeners)))
+    ActorSystem[BakeEvent]("EventBus", Props(behaviours.guardian(eventListeners)))
   }
   implicit val eventBus = new ActorSystemWrapper(eventBusActorSystem)
 
@@ -79,10 +86,10 @@ class AppComponents(context: Context)
   val prism = new Prism(wsClient)
 
   val rootController = new RootController(googleAuthConfig)
-  val baseImageController = new BaseImageController(googleAuthConfig, messagesApi)
+  val baseImageController = new BaseImageController(googleAuthConfig, messagesApi, baseImages)
   val roleController = new RoleController(googleAuthConfig)
-  val recipeController = new RecipeController(googleAuthConfig, messagesApi)
-  val bakeController = new BakeController(eventsSource, prism, googleAuthConfig, messagesApi)
+  val recipeController = new RecipeController(googleAuthConfig, messagesApi, recipes, bakes, baseImages)
+  val bakeController = new BakeController(eventsSource, prism, googleAuthConfig, messagesApi, recipes, bakes, bakeLogs)
   val authController = new Auth(googleAuthConfig)(wsClient)
   val assets = new controllers.Assets(httpErrorHandler)
   lazy val router: Router = new Routes(
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -8,7 +8,9 @@ import com.amazonaws.regions.Regions
 import com.amazonaws.services.dynamodbv2.{ AmazonDynamoDB, AmazonDynamoDBClient }
 import com.gu.cm.{ ConfigurationLoader, Identity }
 import com.gu.googleauth.GoogleAuthConfig
+import com.gu.scanamo.Table
 import data._
+import models.{ Bake, BakeLog, BaseImage, Recipe }
 import org.joda.time.Duration
 import play.api.ApplicationLoader.Context
 import play.api.libs.streams.Streams
@@ -18,7 +20,6 @@ import play.api.libs.iteratee.Concurrent
 import play.api.libs.ws.ahc.AhcWSComponents
 import play.api.routing.Router
 
-import data.Dynamo
 import prism.Prism
 import packer.PackerConfig
 import event.{ ActorSystemWrapper, Behaviours, BakeEvent }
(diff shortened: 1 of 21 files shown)


app/components/AppComponents.scala
```diff
@@ -45,21 +47,30 @@ class AppComponents(context: Context)
...
-    val dynamoClient: AmazonDynamoDBClient = new AmazonDynamoDBClient(awsCreds).withRegion(region)
-    new Dynamo(dynamoClient, identity.stage)
+    val region = configuration.getString("aws.region").map(Regions.fromName).getOrElse(Regions.EU_WEST_1)
+    val dynamoClient: AmazonDynamoDB = new AmazonDynamoDBClient(awsCreds).withRegion(region)
+    new Dynamo(dynamoClient, identity)
   }
   dynamo.initTables()
 
+  def tableName(suffix: String) = Dynamo.tableName(identity, suffix)
+  import DynamoFormats._
```
cb372: Why is this import needed here?

philwills: The `Table` constructor requires that it's type parameter has a `DynamoFormat` in scope.

app/data/BakeLogs.scala
```diff
@@ -1,35 +1,28 @@
...
+import com.gu.cm.Identity
+import com.gu.scanamo.error.DynamoReadError
+import com.gu.scanamo.ops.ScanamoOps
+import com.gu.scanamo.{ Scanamo, ScanamoFree, Table }
+import models._
 
 import scala.collection.JavaConverters._
 
-object BakeLogs {
+class BakeLogs(table: Table[BakeLog]) {
```
cb372: Why not inject `dynamo` into the DAO classes instead of the controllers? Then you can do all the `dynamo.exec` calls in here instead of polluting the controllers with them.

I'd be happiest if the DAO classes acted as DDD-style repositories, returning only domain-level objects (e.g. a `listBakes()` method should return a `List[Bake]` or whatever), and the controllers knew nothing about `Dynamo`, `ScanamoOps` or any other DB-related details.

philwills: My thinking was to push the `ScanamoOps` up to the edge in the same way you would a `Future`, so it instead of action.async or the like, you'd have action.ops or something along those lines. The controller should know how to execute operations, but not need to anything about the structure of the database.

This is the style I'm currently thinking of recommending for Scanamo, but I didn't want to do that without having tried to actually use that style in something approaching a real project like this one.  If you're not happy merging this until I can actually deliver the vision of how I think it can work, then that's completely reasonable.

adamnfish: :+1: for more self-contained interactions with scanamo. Calling a function that returns domain-specific stuff to the controllers is definitely what I'd be after. It'd be in a `Future` of course, since it needs to look up the results.

You're making me nervous with this talk of using an action builder because that seems like the opposite of what @cb372 and I are asking for here. When it returns `Future[MyModel]` I don't need an action builder (or any other internal complexity of scnamo's workings) because Play already provides one and it works for all my stuff, not just the bits that use scanamo.

Unlike @cb372 (I think), I have no problem with a controller-level import that enables this interaction.

philwills:
The `execs` have been banished from the controllers. It now works in the same way as `Action.async`.

The Monad transformers aren't any more necessary than this would be if we were using Futures, but I think make things simpler to work with than without. 

Changing this to make async requests would now be trivial in a way that wouldn't be possible if we executed the I/O directly in the repositories.

app/controllers/BaseImageController.scala
```diff
@@ -1,69 +1,83 @@
...
-          val customisedRoles = ControllerHelpers.parseEnabledRoles(request.body)
-          BaseImages.update(image, description, amiId, customisedRoles, modifiedBy = request.user.fullName)
-          Redirect(routes.BaseImageController.showBaseImage(id)).flashing("info" -> "Successfully updated base image")
-      })
-    }
+  def updateBaseImage(id: BaseImageId) = AuthOpAction(BodyParsers.parse.urlFormEncoded) { implicit request =>
+    (for {
+      image <- OptionT(baseImages.findById(id)).toRight(NotFound)
+      formValues <- XorT.fromXor[ScanamoOps](
+        Forms.editBaseImage.bindFromRequest.toXor.leftMap(formWithErrors =>
```
LATaylor-guardian: It feels like there is allot going on here, is there anyway to make it a little more readable? Maybe by pulling out 

```
XorT.fromXor[ScanamoOps](
 -      }, {		 +        Forms.editBaseImage.bindFromRequest.toXor.leftMap(formWithErrors =>
 -        case (description, amiId) =>		 +          BadRequest(views.html.editBaseImage(image, formWithErrors, Roles.list))))
```
into a function?

app/controllers/RecipeController.scala
```diff
@@ -16,73 +17,83 @@ import scala.util.Try
...
-              val updatedRecipe = Recipes.update(recipe, description, baseImage, customisedRoles, modifiedBy = request.user.fullName, bakeSchedule)
-              bakeScheduler.reschedule(updatedRecipe)
-              Redirect(routes.RecipeController.showRecipe(id)).flashing("info" -> "Successfully updated recipe")
-            case None =>
-              val formWithError = Forms.editRecipe.fill((description, baseImageId, bakeSchedule)).withError("baseImageId", "Unknown base image")
-              BadRequest(views.html.editRecipe(recipe, formWithError, BaseImages.list().toSeq, Roles.list))
-          }
-      })
-    }
+  def updateRecipe(id: RecipeId) = AuthOpAction(BodyParsers.parse.urlFormEncoded) { implicit request =>
```
LATaylor-guardian: As per my previous comment.

app/controllers/RecipeController.scala
```diff
@@ -16,73 +17,83 @@ import scala.util.Try
...
-                val recipe = Recipes.create(id, description, baseImage, customisedRoles, createdBy = request.user.fullName, bakeSchedule)
-                bakeScheduler.reschedule(recipe)
-                Redirect(routes.RecipeController.showRecipe(id)).flashing("info" -> "Successfully created recipe")
-              case None =>
-                val formWithError = Forms.createRecipe.fill((id, description, baseImageId, bakeSchedule)).withError("baseImageId", "Unknown base image")
-                BadRequest(views.html.newRecipe(formWithError, BaseImages.list().toSeq, Roles.list))
-            }
-        }
-    })
+  def createRecipe = AuthOpAction(BodyParsers.parse.urlFormEncoded) { implicit request =>
```
LATaylor-guardian: As per my previous comment.

app/controllers/BaseImageController.scala
philwills: Yea I had the same thought. It's one of those things that almost, comes out as a neat general function, but doesn't quite generalise without massive parameter lists.  Even just pulling it out to a val is probably useful though.

app/controllers/package.scala
```diff
@@ -0,0 +1,8 @@
+import cats.data.Xor
+import play.api.data.Form
+
+package object controllers {
```
LATaylor-guardian: Nice.

app/controllers/package.scala
```diff
@@ -0,0 +1,8 @@
...
+  implicit class FormAsXor[T](form: Form[T]) {
+    def toXor = form.fold(Xor.left, Xor.right)
```
LATaylor-guardian: Nice.

app/controllers/BakeController.scala
```diff
@@ -10,27 +11,39 @@ import play.api.i18n.{ I18nSupport, MessagesApi }
...
 
-  def startBaking(recipeId: RecipeId) = AuthAction { request =>
-    Recipes.findById(recipeId).fold[Result](NotFound) { recipe =>
-      val buildNumber = Recipes.incrementAndGetBuildNumber(recipe.id).get
-      val theBake = Bakes.create(recipe, buildNumber, startedBy = request.user.fullName)
-      PackerRunner.createImage(theBake, prism, eventBus)
+  def startBaking(recipeId: RecipeId) = AuthOpAction { request =>
+    (for {
+      recipe <- OptionT(recipes.findById(recipeId))
+      buildNumber = recipes.incrementAndGetBuildNumber(recipe.id).get
```
cb372: Wow, I can't believe I wrote an `Option.get`!! For shame. Would you mind fixing it is part of this PR?

LATaylor-guardian: Quick, clean the history!

adamnfish: https://help.github.com/articles/deleting-your-user-account/

app/schedule/ScheduledBakeRunner.scala
```diff
@@ -7,20 +7,20 @@ import packer.{ PackerRunner, PackerConfig }
...
-      Recipes.findById(recipeId) match {
+      dynamo.exec(recipes.findById(recipeId)) match {
         case Some(recipe) =>
           // sanity check: is the recipe actually scheduled?
           if (recipe.bakeSchedule.isEmpty) {
             Logger.warn(s"Skipping scheduled bake of recipe $recipeId because it does not have a bake schedule defined")
           } else {
-            val buildNumber = Recipes.incrementAndGetBuildNumber(recipe.id).get
-            val theBake = Bakes.create(recipe, buildNumber, startedBy = "scheduler")
+            val buildNumber = recipes.incrementAndGetBuildNumber(recipe.id).get
```
philwills: 🏃 ✂️

app/controllers/BaseImageController.scala
```diff
@@ -1,69 +1,83 @@
...
+          BadRequest(views.html.editBaseImage(image, formWithErrors, Roles.list))))
+      (description, amiId) = formValues
+      customisedRoles = ControllerHelpers.parseEnabledRoles(request.body)
+      _ <- xorTright(
+        baseImages.update(image, description, amiId, customisedRoles, modifiedBy = request.user.fullName))
+    } yield {
+      Redirect(routes.BaseImageController.showBaseImage(id)).flashing("info" -> "Successfully updated base image")
+    }).merge
```
cb372: Sorry, I'm trying really hard to see how this code

```scala
    (for {
      image <- OptionT(baseImages.findById(id)).toRight(NotFound)
      formValues <- XorT.fromXor[ScanamoOps](
        Forms.editBaseImage.bindFromRequest.toXor.leftMap(formWithErrors =>
          BadRequest(views.html.editBaseImage(image, formWithErrors, Roles.list))))
      (description, amiId) = formValues
      customisedRoles = ControllerHelpers.parseEnabledRoles(request.body)
      _ <- xorTright(
        baseImages.update(image, description, amiId, customisedRoles, modifiedBy = request.user.fullName))
    } yield {
      Redirect(routes.BaseImageController.showBaseImage(id)).flashing("info" -> "Successfully updated base image")
    }).merge
```

is better than this code

```scala
    BaseImages.findById(id).fold[Result](NotFound) { image =>
      Forms.editBaseImage.bindFromRequest.fold({ formWithErrors =>
        BadRequest(views.html.editBaseImage(image, formWithErrors, Roles.list))
      }, {
        case (description, amiId) =>
          val customisedRoles = ControllerHelpers.parseEnabledRoles(request.body)
          BaseImages.update(image, description, amiId, customisedRoles, modifiedBy = request.user.fullName)
          Redirect(routes.BaseImageController.showBaseImage(id)).flashing("info" -> "Successfully updated base image")
      })
    }
```

But to me it looks like the logic of the code is lost in a sea of monad transformers. Is there some kind of Kool-Aid I should be drinking?

I guess my problem is that absolutely everything the controller does now has to be wrapped in `ScanamoOps`, even things unrelated to the DB, like binding the form.

adamnfish: Yeah, I agree. Mentioned to Phil earlier I think it's nice to be aware that's possible (I'm sure there are use cases where scanamo-ing the controller comes out nicely) but I am ill-satisfied with the `AuthOpAction` approach in general. This controller is a great example of why.

I feel like Scanamo should be providing monads that the application can bend to its will, rather than the other way round. Would it be better to convert that Scanamo response to something that fits your controller elsewhere in the application so you only deal with sensible domain types here?

**Disclaimer:** I understand this a lot less well than you two.

philwills: You don't have to bind the form wrapped in a `ScanamoOps`, perhaps I should try a more literal translation for a better comparison. 

I like the flow of everything having a consistent shape where you're either in a `Right` with a value that eventually gets translated into a `Result` or a left with some form of error `Result`, but it does impose a cost of lifting some values which don't start in that shape into it.

I might try and doing this with a `Future` directly for comparison, but I'm also content if you want to close the PR and upgrade or not later. I've learnt a lot form the exercise.

cb372:
Superseded by #55

PR #33, opened 2016-05-24
Additional ami tags

adamnfish:
@cb372 Adds the tags like we discussed. Lemme know if you think this implementation makes sense?


cb372:
👍 Welcome to the team!

PR #34, opened 2016-05-25
ensure checkboxes don't jump around

Reettaphant:
Ensure that checkboxes don't jump around the page when editing base images or recipes.


cb372:
Thank you! I haven't tested it but I'll trust you that this works. 👍

PR #36, opened 2016-05-26
Port the core machine-images functionality to Ansible roles

cb372:
* New roles to install some useful packages, set up NTP, install AWS tools, install the enhanced networking driver
* Add a dependency on the `apt` role to any roles that install packages
* Make Ansible verbose so we can see what on earth it's doing
* Improvements to Packer log parsing to handle the verbose Ansible logs
* Rename `ubuntu-wily-init` role to plain old `ubuntu-init`, because it doesn't do anything Wily-specific


cb372:
@Reettaphant Sorry, this PR probably conflicts with the work that you're doing. Hopefully the merge conflict will not be too bad.

app/models/BakeLog.scala
```diff
@@ -7,7 +7,7 @@ case class MessagePart(text: String, colour: String) {
 
   def toHtml = {
     val htmlColour = MessagePart.HtmlColours.getOrElse(colour, colour)
-    s"""<span style="color: $htmlColour">$text</span>"""
+    s"""<span style="color: $htmlColour">${text.replaceAllLiterally("\n", "<br>")}</span>"""
```
piuccio: or you could have used `<pre>` instead of `<span>`

cb372:
@LATaylor-guardian Once this is merged, AMIgo images will be officially ready for prime time!

@adamnfish @sihil might want to review.

piuccio:
👍

LATaylor-guardian:
@cb372 Great! Can try tomorrow.

app/models/BakeLog.scala
cb372: True. There's an open issue to make the page that show the bake logs look less terrible, so I'll leave it for now.

roles/useful-packages/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+---
```
adamnfish: I'm unconvinced stuff like this should be included in bakes rather than it being a "install this stuff manually if you ever need to `ssh` onto an instance for some weird reason" kind of thing. There's no reason these would be needed outside of something application-specific and in those cases you'd probably only one or two.

cb372: So my commit message `New role to install some uncontroversial packages` was a little optimistic!

Some of these packages are useful for things that run in UserData or cron jobs for a lot of apps:

* `language-pack-en`: not sure what it is exactly, but it sounds important
* `cloud-guest-utils` provides the `ec2metadata` command, which is used in various places by my team's software
* `unzip`: I'm sure a lot of people want to download stuff from S3 and then unzip it in UserData
* `jq`: this is used by a lot of scripts that call AWS APIs and need to parse the responses. I think removing it would break a lot of stuff.

Apart from that:
* `wget` and `python-setuptools` are already installed anyway, so I'll get rid of it from here
* `build-essential` installs `make` and its brethren, maybe not needed by anyone. Actually it's already installed by the `enhanced-networking` role anyway. I'll get rid of it.

How does that sound?

adamnfish: Yeah, `language-pack-en` is part of setting up the localisation which is needed to do basically anything without lots of warnings.

It was `jq` that triggered my lookings and then I was particularly chatting about stuff like `python-setuptools`, `build-essential` and `unzip`, but I take your point about userdata stuff. I guess the Python one's needed for `pip` which is needed for the aws cli - reasonable to expect.

Fair enough!

PR #39, opened 2016-06-03
Add roles for grid

Reettaphant:
- Add roles needed for cropper and imgops in grid @kenoir

### DIFF
--- a/roles/imgops-logrotate/tasks/main.yml
+++ b/roles/imgops-logrotate/tasks/main.yml
@@ -1,0 +1,3 @@
+---
+- name: Move imgops config files into logrotate folder
+  copy: src=logrotate dest=/etc/logrotate.d/imgops owner=root group=root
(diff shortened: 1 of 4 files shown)


roles/imgops-logrotate/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Move imgops config files into logrotate folder
+  copy: src=logrotate dest=/etc/logrotate.d/imgops
```
cb372: You can set the owner and group as part of the `copy`, meaning you can delete the `Change owner to root` task.

```
copy: src=logrotate dest=/etc/logrotate.d/imgops owner=root group=root
```

cb372:
Looks good apart from one small simplification

Reettaphant:
I've made that change now, thanks!

cb372:
👍  I'll deploy Amigo with these changes now

PR #40, opened 2016-06-06
Add a role to install Logstash

cb372:
CAPI still needs Logstash to send non-application logs (GC logs, Elasticsearch slowlog, etc.)

This role has been tested in DEV and seems to work as expected, but may need minor tweaks after I try it with the CAPI apps.


mchv:
👍

PR #44, opened 2016-06-28
[WIP] Adds OpenResty setup for ImgOps

kenoir:
OpenResty: https://openresty.org/en/

This allows us to add Nginx modules that facilitate authentication with
S3 from Nginx.


kenoir:
Closing as for some reason the image_filter module fails to work correctly when built like this.

PR #46, opened 2016-08-22
Update Kong to 0.9.0 to support the use of a Postgres database.

LATaylor-guardian:


cb372:
Looks like we would need to pass in an nginx conf template when starting Kong if we want to enable access logging.

https://github.com/Mashape/getkong.org/pull/263/files#diff-b1e96bff345e56d19bb327977c2a7198R120

LATaylor-guardian:
@cb372 Will do this now.

roles/kong/files/kong.logrotate
```diff
@@ -8,6 +8,6 @@
 	notifempty
 	nocreate
 	postrotate
-		[ ! -f /usr/local/kong/kong.pid ] || kill -USR1 `cat /usr/local/kong/kong.pid`
+		[ ! -f /usr/local/kong/nginx.pid ] || kill -USR1 `cat /usr/local/kong/nginx.pid`
```
cb372: We'll have to test whether this works with Kong 0.9

roles/kong/files/configure.sh
```diff
@@ -2,30 +2,40 @@
...
 # Also sets Kong's cluster advertise address to the instance's internal IPv4 address
 #
 # This script must be run as root
 set -e
 
-USAGE="Usage: $0 [cassandrahost1 [cassandrahost2 ...]]
+USAGE="Usage: $0 postgres_host postgres_username postgres_password
 
-Example: $0 1.2.3.4 5.6.7.8
+Example: $0 1.2.3.4 kong mysecretpassword
```
cb372: This example has a confusing value for `postgres_host`, and `postgres_username` is missing

roles/kong/files/configure.sh
```diff
@@ -2,30 +2,40 @@
...
 "
+test -z $1 && echo 'Postgres host missing.' && exit 1
```
cb372: You can write this more simply as `replace_postgres_host=${1?Postgres_host_missing}`

roles/kong/files/configure.sh
```diff
@@ -2,30 +2,40 @@
...
-  replace_cass_hosts="$replace_cass_hosts    - \"${host}:9042\"\n"
-done
+search_postgres_host='POSTGRES_HOST'
+replace_postgres_host=$1
+
+search_postgres_username='POSTGRES_USERNAME'
+replace_postgres_username=$2
+
+search_postgres_password='POSTGRES_PASSWORD'
+replace_postgres_password=$3
```
cb372: Passing the password as a command line argument is a bit dodgy in general, as it might remain in bash history, etc. Maybe best to write it to a file, pass the filename to the script, then delete the file after it has run.

A couple of related security issues I noticed:
* Kong's config file is world-readable. Not much point hiding the password from the bash history if it's going to be written to a world-readable file! ;)
* Kong is running as root. We should create a dedicated user and use that user to start Kong.

roles/kong/files/configure.sh
LATaylor-guardian: postgres_username is `kong`

cb372:
Looking more closely, maybe we don't need to customise the nginx config? It looks like we got lucky and access logging is enabled by default in Kong 0.9.

https://github.com/Mashape/kong/blob/0.9.0/kong/templates/nginx_kong.lua#L6

LATaylor-guardian:
@cb372 https://github.com/Mashape/kong/blob/0.9.0/kong/templates/nginx_kong.lua#L135

cb372:
Looks like a good start. I'll try building an AMI locally so we can test in CODE tomorrow.

PR #53, opened 2016-09-14
Add a role for regularly downloading SSH keys

philwills:


philwills:
Note that this leaves specifying the bucket to recipe configuration, so isn't dependent on the changes @sihil, @philmcmahon and I discussed yesterday.

roles/github-ssh-keys/templates/install-keys.sh.template
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+aws s3 cp s3://{{github_keys_bucket}}/{{github_team_name}}/authorized_keys /tmp/authorized_keys
+
+cp /tmp/authorized_keys /home/{{ssh_user}}/.ssh/authorized_keys
+
+chmod 600 /home/{{ssh_user}}/.ssh/authorized_keys
```
sihil: these `cp` and `chmod` commands could be combined into a single `install` command

sihil:
Can there be a README with some details in such as how to check that this is running / has run and any configuration notes?

roles/github-ssh-keys/files/update-keys.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update SSH keys every half hour
+
+[Timer]
+OnBootSec=30 seconds
+OnUnitActiveSec=30 minutes
+
+[Install]
+WantedBy=timers.target
```
sihil: This is really nice. Can very much see this replacing cron completely.

roles/github-ssh-keys/templates/install-keys.sh.template
philwills: Ooh, fancy, I'll give it a go.

philwills:
Yea, a README sounds good.

philwills:
Something that occurred to me is that this has nothing to do with Github, so I'm tempted to rename to S3 SSH keys and the parameters:

`github_keys_bucket` -> `ssh_keys_bucket`
`github_team_name` -> `ssh_keys_prefix`

I can see good arguments both ways on this, so would welcome feedback.

sihil:
I would totally advocate for making it generic and simply documenting the expected format of the bucket.

philmcmahon:
Yep, that sounds sensible.

roles/s3-ssh-keys/files/update-keys.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update SSH keys every half hour
+
+[Timer]
+OnStartupSec=1 minute
```
philwills: I tried making this sooner, but it seems that if you try to do it too quickly the aws instance profile credentials are unavailable.

sihil:
Perfect. 👍

philmcmahon:
Brilliant! 🔐

PR #54, opened 2016-09-15
Add `iam:PassRole` to packer permissions

philwills:
This allows Packer to start a machine with an instance profile. See the bottom of https://www.packer.io/docs/builders/amazon.html


philwills:
This was preventing all bakes from working, so I've already applied this and this is FYI.

PR #56, opened 2016-09-20
Upgrade configuration-magic

cb372:
This fixes an AWS SDK version conflict caused by the Scanamo upgrade. Configuration-magic doesn't make any calls to the EC2 API when you're running on your local machine, so the problem only manifested itself in production. My favourite kind of bug.


LATaylor-guardian:
👍

PR #62, opened 2016-09-27
Add a housekeeping script for Kong

cb372:
This script runs as a nightly cron job and deletes old entries from one of Kong's DB tables. Without it, the table will grow indefinitely until the DB machine fills its disk.

Also install the `psql` command on the Kong machine.

This has been tested in CODE.


LATaylor-guardian:
👍 Once build passes.

PR #64, opened 2016-10-26
Adds role to install pluto

kenoir:


fredex42:
please keep me updated as progress on this as well, working with @kenoir

LATaylor-guardian:
@kenoir build is failing.

cb372:
>[error] (*:riffRaffUpload) com.amazonaws.AmazonClientException: Unable to load AWS credentials from any provider in the chain

I set up Travis and I might have used my own IAM user's AWS creds. Doesn't sound like something I would do, but anyway... You just need to create an IAM user for Travis and update the `AWS_*` environment variables in the Travis UI.

fredex42:
@LATaylor-guardian do you know which AWS account we need credentials for here? Can one of us do it or does @cb372 need to set up the IAM user?

LATaylor-guardian:
@fredex42 I think this resides in the deploy tools account which I don't have access too.

fredex42:
@adamnfish could you sort out an IAM user so we can run the amigo tests again please?

adamnfish:
I can, pop by my desk. Otherwise maybe better to get access via Janus so you can manage this yourselves.

kenoir:
Closing as no need to use fork now.

PR #66, opened 2016-10-28
Make s3-ssh-keys redhat compatible

philmcmahon:
Getting pip3 installed on RedHat is a pain so I used pip instead. Now the aws-tools script does different stuff based on whether its on RedHat or Debian.

Ansible has a dependency on libselinux, which doesn't come pre installed on redhat, so I added a role to do that, which is now a dependency of s3-ssh keys when running on RedHat.

### DIFF
--- a/roles/aws-tools/tasks/install-aws-cli-redhat6.yml
+++ b/roles/aws-tools/tasks/install-aws-cli-redhat6.yml
@@ -1,0 +1,12 @@
+---
+- name: Enable IUS (ius.io) packages on RedHat
+  yum: name=https://centos6.iuscommunity.org/ius-release.rpm state=present
+
+- name: Install python3 on centos
+  yum: name=python35u state=present
+
+- name: Install pip on centos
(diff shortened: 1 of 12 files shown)


roles/aws-tools/tasks/install-aws-cli-redhat6.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Enable IUS (ius.io) packages on RedHat
+  yum: name=https://centos6.iuscommunity.org/ius-release.rpm state=present
+
+- name: Install python3 on centos
+  yum: name=python35u state=present
```
fredex42: just thinking, if you're installing python-pip that's presumably using python 2.x so is python3 required?

kenoir: I think this will work without python3, even though the AWS docs recommend it.

kenoir:
:+1:

PR #67, opened 2016-10-31
Add the concept of Linux distribution

kenoir:
In order to provision redhat boxes for Pluto.

screenshot from 2016-10-31 14-15-58

Will default to Ubuntu when not identified or creating a new (or modifying an old) base image.

### DIFF
--- a/app/packer/PackerBuildConfigGenerator.scala
+++ b/app/packer/PackerBuildConfigGenerator.scala
@@ -1,9 +1,9 @@
 package packer
 
-import java.nio.file.{ Paths, Path }
+import java.nio.file.{Path, Paths}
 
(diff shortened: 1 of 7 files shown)


app/packer/PackerBuildConfigGenerator.scala
```diff
@@ -30,7 +30,11 @@ object PackerBuildConfigGenerator {
       subnet_id = packerConfig.subnetId,
       source_ami = "{{user `base_image_ami_id`}}",
       instance_type = "t2.micro",
-      ssh_username = "ubuntu",
+
+      ssh_username = bake.recipe.baseImage.linuxDist
+        .map(_.loginName)
+        .getOrElse("ec2-user"),
```
jennysivapalan: I think we should fall back to `ubuntu` user as that is default for most cases.

jennysivapalan:
👍

PR #69, opened 2016-11-02
Amigo - modify kong truncate-ratelimiting_metrics script.

LATaylor-guardian:
- Run truncate script every two hours rather than every 24 hours to reduce impact on Postgres.
 - Rewrite query to make it clearer how to limit a delete query.

see: http://docs.ansible.com/ansible/cron_module.html for ansible cron docs.

### DIFF
--- a/roles/kong/files/truncate-ratelimiting_metrics.sh
+++ b/roles/kong/files/truncate-ratelimiting_metrics.sh
@@ -2,6 +2,8 @@
 #
 # Removes old entries from Kong's ratelimiting_metrics table
 # to prevent the table from growing indefinitely.
+# Intended to run periodically throughout the day as opposed to once. This avoids deleting a really large number
+# of rows from Postgres all at once and causing issues with Postgres.
 #
 # Looks up the DB host and password from /etc/kong.conf.
 #
@@ -12,3 +14,5 @@ host=$(grep -o -P '(?<=pg_host = )(.*?)(?= )' /etc/kong.conf)
 password=$(grep -o -P '(?<=pg_password = )(.*?)(?= )' /etc/kong.conf)
 
 PGPASSWORD=$password psql -U kong -h $host -c "delete from ratelimiting_metrics where period_date < now() - '24 hours'::interval"
+
+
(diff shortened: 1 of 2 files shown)


mchv:
👍

roles/kong/files/truncate-ratelimiting_metrics.sh
```diff
@@ -11,4 +13,6 @@
 host=$(grep -o -P '(?<=pg_host = )(.*?)(?= )' /etc/kong.conf)
 password=$(grep -o -P '(?<=pg_password = )(.*?)(?= )' /etc/kong.conf)
 
-PGPASSWORD=$password psql -U kong -h $host -c "delete from ratelimiting_metrics where period_date < now() - '24 hours'::interval"
+PGPASSWORD=$password psql -U kong -h $host -c "delete from ratelimiting_metrics where ctid in (select ctid from ratelimiting_metrics where period_date < now() - '24 hours'::interval)"
```
cb372: Sorry to lurk, but what's the point of this query change? It looks like you're doing the same query as before, just written a slightly different way.

LATaylor-guardian:
Will create a new image off this, update Kong instances and deploy tomorrow.

roles/kong/files/truncate-ratelimiting_metrics.sh
LATaylor-guardian: That's alright. A little background. This script wasn't running on the PROD instances as the AMI hadn't been updated. When it was updated and this script run it nearly killed the DB because there was so many rows to delete. Therefore I had to manually lower the number of items to delete on the DB. To do so you have to limit a delete query  - this isn't obvious how to do this. I figured out how to do it as such: `delete from ratelimiting_metrics where ctid in (select ctid from ratelimiting_metrics where period_date < now() - '24 hours'::interval LIMIT 10000)`

Therefore I wanted to change the query to make this more evident. but in effect there is no actual difference to what is being run.

mchv: @LATaylor-guardian is the `LIMIT 10000` not missing then?

LATaylor-guardian: @mchv No, if you limit the query you could end up falling behind because you're only deleting 10000 and you might have created 15000 rows. Therefore the solution for now is to run the script more often (every two hours) so it has less to delete each time.

PR #70, opened 2016-11-02
When running on RedHat, create cronjob files instead of a systemd process to update keys.

philmcmahon:
As redhat6, which pluto runs on, doesn't have systemd. The cronjob version will need to be kicked off after the instance comes up which isn't as nice as systemd but does the job.


fredex42:
👍 from me

sihil:
Out of interest, how hard have we tried to run pluto on a more modern linux? RedHat 6 was EOL in May this year (well, it's complex - but the "Production 1" date has passed). It's also pretty expensive, although I guess that is a sunk cost for the time being?

kenoir:
@sihil we're following the software providers "recommendations" here, so our hands are a bit tied.

PR #74, opened 2016-11-04
New role that disables missed daily tasks at boot time

sihil:
With Xenial there is a task that does apt related tasks at boot time. This can mean that attempts to install packages at boot time can fail resulting in instances not coming up. See this for some more details: http://stackoverflow.com/questions/36896806/how-can-i-be-sure-a-freshly-started-vm-is-ready-for-provisioning

This PR adds a new role that disables this boot time behaviour so that provisioning of instances never fails (because of this at least).

The intention is to add this to the base Xenial image to avoid surprising people.


jennysivapalan:
👍 😮

PR #78, opened 2016-11-10
Filebeat -> Logstash -> Kinesis

kenoir:
In order to ship logs from Pluto to the editorial-tools ELK stack.

### DIFF
--- a/roles/java8/tasks/main.yml
+++ b/roles/java8/tasks/main.yml
@@ -4,6 +4,11 @@
   with_items:
     - openjdk-8-jre-headless
     - openjdk-8-jdk
+  when: ansible_os_family == "Debian"
+
+- name: Install Java 8
+  yum: name=java state=latest
+  when: ansible_os_family == "RedHat"
 
 ## Workaround for Debian Java packaging bug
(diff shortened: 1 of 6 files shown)


roles/java8/tasks/main.yml
```diff
@@ -12,3 +17,4 @@
 ## https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775775
 - name: Install Java certificates
   command: /var/lib/dpkg/info/ca-certificates-java.postinst configure
+  when: ansible_os_family == "Debian"
```
kenoir: I reckon it's likely as these are to account for a debian bug.

roles/java8/tasks/main.yml
```diff
@@ -4,6 +4,11 @@
   with_items:
     - openjdk-8-jre-headless
     - openjdk-8-jdk
+  when: ansible_os_family == "Debian"
+
+- name: Install Java 8
+  yum: name=java state=latest
```
kenoir: Yup, want to try this out right now, but will update it in another PR.

PR #79, opened 2016-11-10
adding logging bug patch

fredex42:
@kenoir 
Portal has an annoying bug, whereby if a plugin fails to start up properly because of an exception the actual error is not shown in the log, but a generic one about "model User does not exist."

This introduces a patch which prevents the secondary exception and thereby allows the actual problem to be logged.  I use it for all testing and it is in place on the FourthPortalDev image that we use.


kenoir:
:+1:

PR #81, opened 2016-11-10
More ssh keys tidying

philmcmahon:
Missed these off


philwills:
@philmcmahon Can we have a chat about this. We have a systemd timer, so don't need a cron job too.

philwills:
Ah, just seen the other one was Redhat only. I hadn't realised we were using redhat based AMIs for anything Amigo built.

PR #83, opened 2016-11-11
Add role for generically installing an apt package

philwills:
This was driven by wanting `pip` pre-installed to avoid dpkg lock contention with cloud watch logs, but it seems like this would be generally useful.

Not at all sure about the name, so would love alternative suggestions.

### DIFF
--- a/app/ansible/PlaybookGenerator.scala
+++ b/app/ansible/PlaybookGenerator.scala
@@ -20,7 +20,7 @@ object PlaybookGenerator {
     if (role.variables.isEmpty)
       role.roleId.value
     else
-      s"{ role: ${role.roleId.value}, ${role.variables.map { case (k, v) => s"$k: '$v'" }.mkString(", ")} }"
+      s"{ role: ${role.roleId.value}, ${role.variables.map { case (k, v) => s"$k: ${v.quoted}" }.mkString(", ")} }"
   }
 
 }
--- a/app/data/Recipes.scala
+++ b/app/data/Recipes.scala
@@ -3,7 +3,6 @@ package data
 import com.amazonaws.services.dynamodbv2.model._
 import models._
 import org.joda.time.DateTime
-
 import com.gu.scanamo.syntax._
 
 import scala.collection.JavaConverters._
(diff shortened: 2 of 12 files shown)


davidfurey:
Great idea.  Can you use this to install multiple packages?

philwills:
@davidfurey Good question, I'll have a look and see. Would seem silly not to be able to.

sihil:
I'm not entirely against this, but it seems like it could be subject to prolific undocumented use. this doesn't enable any new abilities (we can achieve this with a role that, to some extent, self-documents intent).

Is this really saying that roles are too difficult to add?

philmcmahon:
Could you make it work with yum? :trollface:

kenoir:
+1 what @sihil said. I think it would be worth putting work into extracting the roles from the code and making them easier to modify (i.e. not requiring a deploy).

I've created an issue for this: https://github.com/guardian/amigo/issues/86

philwills:
@sihil I think they're enough of a pain that it's reasonable to avoid for the simplest cases, e.g. install this one package. Even for someone like me who's used to the process, that would be an hour or two  of effort plus considerably more elapsed time, compared to 5 minutes with this available.

I think `apt-install` with a a parameter of `python-pip` is no less self-documenting than a role called, `pip (deb-based distros)`

sihil:
Except that's doesn't seem like a good role name IMHO. A role is about the task you want to use it for. I guess the recipe name is useful here and each recipe could potentially have a notes or description field too.

philwills:
Also, I'm really appreciative of all the feedback and discussion of this.

philwills:
@sihil I'm not sure what you mean. I don't understand how the existing roles fit that description.

sihil:
Yeah - fair enough - I've just reviewed them and there isn't really anything special there. I guess my concern is that people eventually stop doing roles entirely by creating a class of roles that let them install packages, files, etc etc with ever more complex recipes. Maybe that's OK, but I'd rather make roles easier to edit.

I propose that we do this and see what happens.

philwills:
I've updated the task with a more declarative name, which suggested itself when adding support for RedHat derivatives. I've also added support for installing multiple packages which has meant introducing a small parser to replace a regular expression.

philmcmahon:
Brilliant! Code looks good to me 👍

PR #87, opened 2016-11-15
Flexible Mongo3

Reettaphant:
Add roles for flexible secondary to use mongo 3.20

@jennysivapalan


roles/disable-transparent-huge-pages/files/hugepages
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
philmcmahon: Tiny thing but this file should have a .sh prefix

roles/disable-transparent-huge-pages/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+  dependencies:
+    - mongo-helper-scripts
+    - mongo30
```
philmcmahon: I'm not sure about this, maybe it should be the other way round? So mongo depends on this? ...or maybe the  the different roles could be independent?

roles/usage-monitoring-agent/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+  dependencies:
+    - mongo30
```
philmcmahon: Why do we need this dependency? This feels like something that would be useful to have as a generic role rather than being dependent on mongo.

Reettaphant:
@philmcmahon I've now updated the depedencies for these and separated installing aws-sdk into its own role away from mongo helper scripts

roles/install-rsyslog-config/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: install rsyslog config
+  shell: |
+      cat > /etc/rsyslog.d/31-mongo-scripts.conf <<EOF
```
philmcmahon: could this be part of the mongo script since its mongo specific? Sorry to be picky!

roles/install-sysfsutils/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: install sysfsutils
```
jennysivapalan: This should be with mongo30

jennysivapalan:
👍

PR #89, opened 2016-11-18
who wouldn't want a big ulimit?

johnduffell:
We are hitting ulimits on membership.  Since we are all running single purpose machines, I doubt everyone would complain about one user opening lots of files.  And the jvm does like to open lots of files...

@rtyley

### DIFF
--- a/roles/kong/tasks/main.yml
+++ b/roles/kong/tasks/main.yml
@@ -24,13 +24,6 @@
 - name: Configure Kong log rotation
   copy: src=kong.logrotate dest=/etc/logrotate.d/kong
 
-- name: Bump ulimits
-  blockinfile:
-    dest: /etc/security/limits.conf
-    block: |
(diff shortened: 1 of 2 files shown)


rtyley:
Here's an example of the exceptions we're getting - they say "Too many open files":

image

https://sentry.io/the-guardian/membership/issues/182649206/

johnduffell:
@LATaylor-guardian will that cause any issue moving the ulimit for kong?

LATaylor-guardian:
@johnduffell I don't believe so as the Kong image uses ubuntu-init as a base image so it should still pick this up.

rtyley:
:+1: - the CI build seems to be using bad credentials for RiffRaff upload tho.

johnduffell:
opened a new one from a branch instead of a fork https://github.com/guardian/amigo/issues/90

PR #90, opened 2016-11-18
who wouldn't want a big ulimit?

johnduffell:
this time with the branch in the same repo

All as per https://github.com/guardian/amigo/pull/89

@rtyley @LATaylor-guardian

### DIFF
--- a/roles/ubuntu-init/tasks/main.yml
+++ b/roles/ubuntu-init/tasks/main.yml
@@ -4,3 +4,10 @@
 
 - name: Configure locale
   command: locale-gen en_GB.UTF-8
+
+- name: Bump ulimits
+  blockinfile:
+    dest: /etc/security/limits.conf
+    block: |
+      *  soft  nofile  16384
+      *  hard  nofile  16384
--- a/roles/kong/tasks/main.yml
+++ b/roles/kong/tasks/main.yml
@@ -24,13 +24,6 @@
 - name: Configure Kong log rotation
   copy: src=kong.logrotate dest=/etc/logrotate.d/kong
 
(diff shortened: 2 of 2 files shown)


rtyley:
:+1: using a branch rather than a fork means that the riffraff credentials are now available, right!

rtyley:
@johnduffell looks like amigo is needs to be manually deployed https://riffraff.gutools.co.uk/deployment/history?projectName=amigo&page=1#

roles/ubuntu-init/tasks/main.yml
```diff
@@ -4,3 +4,10 @@
 
 - name: Configure locale
   command: locale-gen en_GB.UTF-8
+
+- name: Bump ulimits
+  blockinfile:
+    dest: /etc/security/limits.conf
+    block: |
+      *  soft  nofile  16384
```
paulbrown1982: @johnduffell I think this needs to be root and not * for the way we load the apps. the 'frontend' user, a shadow of root I think, has only a limit of 4096 

http://stackoverflow.com/questions/21515463/how-to-increase-maximum-file-open-limit-ulimit-in-ubuntu

johnduffell: Oh dear didn't realise that.

rtyley: Did this lead to https://github.com/guardian/amigo/pull/94 to get the limit set for the frontend user?

johnduffell: in the end it was nothing to do with user limits, actually systemd has service limits that are completely separate, but that's the right PR

PR #94, opened 2016-11-21
make systemd services have a big ulimit

johnduffell:
@philwills @LATaylor-guardian @rtyley hopefully this will give all systemd services a bigger no of open files.


johnduffell:
this seems to work on a test on an existing instance:
`Max open files            65536                65536                files     
`

PR #95, opened 2016-11-21
Move install of rsyslog and sysfsutils to mongo30

jennysivapalan:
As discussed I've moved rsyslog and sysfsutils to mongo30.

@Reettaphant @philmcmahon I can merge this into @Reettaphant branch if you're happy with this.


philmcmahon:
👍

PR #97, opened 2016-12-07
a couple of small tweaks for first time users

jharewinton:
Added packer dependency to the readme for running locally. This caught me and @jennysivapalan out

Also advises that description is mandatory when creating a recipe


jennysivapalan:
👍 📦

PR #98, opened 2016-12-15
Cloudwatch logs role

alexduf:
This allow us to bake images with the cloudwatch-log agent.

It also provide a utility script to reduce the boiler plate in the user-data within the cloudformation.
It assumes the distribution is using systemd

The readme should explain it all, so if it isn't clear please let me know I'll do my best to improve it.

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -42,7 +42,7 @@ class AppComponents(context: Context)
     val awsCreds = new AWSCredentialsProviderChain(
       new ProfileCredentialsProvider("deployTools"),
       new ProfileCredentialsProvider(),
-      new InstanceProfileCredentialsProvider
+      InstanceProfileCredentialsProvider.getInstance()
     )
     val region = configuration.getString("aws.region").map(Regions.fromName).getOrElse(Regions.EU_WEST_1)
     val dynamoClient: AmazonDynamoDBClient = new AmazonDynamoDBClient(awsCreds).withRegion(region)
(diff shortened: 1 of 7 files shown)


app/components/AppComponents.scala
```diff
@@ -42,7 +42,7 @@ class AppComponents(context: Context)
     val awsCreds = new AWSCredentialsProviderChain(
       new ProfileCredentialsProvider("deployTools"),
       new ProfileCredentialsProvider(),
-      new InstanceProfileCredentialsProvider
+      InstanceProfileCredentialsProvider.getInstance()
```
philwills: Why is this change required?

roles/cloud-watch-logs/README.md
```diff
@@ -0,0 +1,47 @@
+This role will install the [cloudwatch](http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html) agent.
+
+It also provide a utility script to configure your logs in `/opt/cloudwatch-logs/configure-logs`
+The arguments are:
+ - an id, to distinguish between log file (for instance `application` and `gc`)
+ - the stack
```
philwills: We could default these from tags, but I'm not sure that's worth the extra complexity.

app/components/AppComponents.scala
alexduf: It actually isn't, but the direct creation of the `InstanceProfileCredentialsProvider` is deprecated now

roles/cloud-watch-logs/README.md
alexduf: I hadn't thought of it, but that would be a great improvement. I'd like to merge it as is though as I don't think I should spend more time on it.

app/components/AppComponents.scala
philwills: Perfect, thanks.

PR #99, opened 2016-12-16
browser testing role

jharewinton:
installs Google Chrome, Chromedriver and the virtual frame buffer, Xvfb, to allow for headless browser testing on fully featured Chrome

### DIFF
--- a/roles/browser-testing/tasks/main.yml
+++ b/roles/browser-testing/tasks/main.yml
@@ -1,0 +1,27 @@
+---
+- name: Install Xvfb
+  apt: name=xvfb state=latest
+
+- name: Download Chrome
+  get_url: url=https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb dest=/tmp/google-chrome-stable_current_amd64.deb
+
+- name: Configure Chrome
+  command: dpkg -i /tmp/google-chrome-stable_current_amd64.deb
+
+- name: Install Chrome missing packages
+  command: apt-get install -f -y
+
+- name: Download Chromedriver
+  get_url: url=http://chromedriver.storage.googleapis.com/2.25/chromedriver_linux64.zip dest=/tmp/chromedriver.zip
+
+- name: Unzip Chromedriver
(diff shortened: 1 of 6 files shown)


roles/browser-testing/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  shell: apt-get install -f -y
+
+- name: Download Chromedriver
+  get_url: url=http://chromedriver.storage.googleapis.com/2.25/chromedriver_linux64.zip dest=/tmp/chromedriver.zip
+
+- name: Unzip Chromedriver
+  unarchive: src=/tmp/chromedriver.zip dest=/usr/local/bin/
+
+- name: Enable Chromedriver
+  shell: chmod 751 /usr/local/bin/chromedriver
```
philmcmahon: Similarly, you could use http://docs.ansible.com/ansible/file_module.html here.

roles/browser-testing/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Download Chrome
+  get_url: url=https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb dest=/tmp/google-chrome-stable_current_amd64.deb
+
+- name: Install Chrome - ignores errors as expect missing packages
+  shell: dpkg -i /tmp/google-chrome-stable_current_amd64.deb
+  ignore_errors: yes
+
+- name: Install missing packages for Chrome
+  shell: apt-get install -f -y
```
jharewinton: the ansible apt package doesn't support this particular apt command. It 'fixes' the missing packages from the chrome install, which vary from version to version of chrome

philmcmahon:
Other than the above comments, LGTM 👍

roles/browser-testing/tasks/main.yml
philwills: Might it be worth using the long option here, to make it more obvious for those of us less familiar with apt-get flags?

PR #100, opened 2016-12-20
add copy buttons for amis on recipes & individual bakes

jharewinton:
big thanks to @ShaunYearStrong for help with this

ended up needing a bit of rework of the tables

screen shot 2016-12-20 at 15 22 11
screen shot 2016-12-20 at 15 22 22


philmcmahon:
👍

philwills:
Wow, the clipboard API actually works these days! Last time I did something like this it was still some hacky flash madness.

PR #103, opened 2017-01-19
Add nginx+pagespeed role. Compile and install nginx with pagespeed module

DiegoVazquezNanini:
Adds a new nginx+pagespeed role.
The role compiles and install nginx with pagespeed module.
@philmcmahon @philwills @TBonnin @jfsoul @rich-nguyen


roles/frontend-router/README.md
```diff
@@ -0,0 +1,9 @@
+# frontend-router
+
+Compile and install Nginx version with the pagespeed module.
+The `frontend-router` parameter specifies which version(s) of nginx and pagespeed to use.
+If no specific version is provided the following values are use.
+
+```
+nginx_version: 1.8.1, ngxpagespeed_version: 1.12.34.2
```
TBonnin: I would not mention the version here but only link to the `default/main.yml` file. Less maintenance and file to modify when upgrading version 😉

roles/frontend-router/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  user: name=nginx group=nginx
+
+- name: Download and compile nginx with pagespeed module
+  shell: wget https://ngxpagespeed.com/install; chmod +x install; ./install -v {{ ngxpagespeed_version }} -n {{ nginx_version }} -y -a '--sbin-path=/usr/sbin/nginx --conf-path=/etc/nginx/nginx.conf'
+
+- name: Create the directory /var/log/nginx
+  file: path=/var/log/nginx state=directory owner=nginx group=nginx
+
+- name: Create the directory /etc/nginx/conf.d
+  file: path=/etc/nginx/conf.d state=directory owner=nginx group=nginx
```
TBonnin: what is this directory used for?

jfsoul: that's where we put `router.conf` - get's included by `/etc/nginx/nginx.conf` IIRC

TBonnin:
Should this role be called `nginx+pagespeed` rather than `frontend-router`? That would describe what it is rather where it is going to be used
What do you think?

roles/frontend-router/tasks/main.yml
TBonnin: Oh yeah of course. Thx

philwills:
I agree with @TBonnin that there isn't actually much router specific in here, so `nginx+pagespeed` seems like a good name.

PR #104, opened 2017-01-19
load image from assets

akash1810:
Currently the sombrero image is being loaded from http://publicdomainvectors.org/photos/1385029480.png which creates a mixed content warning as its not https. Lets just load it from our assets.


akash1810:
closing over #106 because this PR wasn't building, even though now it says it does build 😕

PR #106, opened 2017-01-20
load image from assets... again

akash1810:
I think #104 is failing to build because it is a fork. Raising this PR to test the theory.


akash1810:
merging as #104 was approved

PR #107, opened 2017-01-23
Fix permision on the pid file. Adds more compilation properties

DiegoVazquezNanini:
Small set of fixes for the role

@TBonnin @jfsoul

### DIFF
--- a/roles/nginx+pagespeed/files/nginx.conf
+++ b/roles/nginx+pagespeed/files/nginx.conf
@@ -1,7 +1,7 @@
 user                nginx;
 worker_processes    auto;
 error_log           /var/log/nginx/error.log;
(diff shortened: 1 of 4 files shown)


roles/nginx+pagespeed/files/nginx.conf
```diff
@@ -1,7 +1,7 @@
 user                nginx;
 worker_processes    auto;
 error_log           /var/log/nginx/error.log;
-pid                 /var/run/nginx.pid;
+pid                 /var/run/nginx/nginx.pid;
```
TBonnin: why this change?

DiegoVazquezNanini:
I do need to log file configs con the compilation and I also need to remove the nginx users from the systemd definition.

PR #108, opened 2017-01-24
Allow forward slash in parameter values.

susiecoleman:
Updated as part of work to write a role to allow installation of elasticsearch plugins. It needs to be possible to have `/` in a parameter value. E.g. `plugins:elasticsearch/elasticsearch-cloud-aws/2.7.0`


jennysivapalan:
👍 ❗️

PR #109, opened 2017-01-24
Allow parameter values to have a dot.

susiecoleman:
Follow up from this [PR](https://github.com/guardian/amigo/pull/108). Elasticsearch plugins can have `.` in them so it needs to be possible to specify parameter values with a `.` in order to write an elasticsearch install plugins role.


jennysivapalan:
👍

PR #111, opened 2017-01-26
Include roles for elk stack

susiecoleman:


jharewinton:
👍

susiecoleman:
Opened too early, still work to do on this

Issue #28, opened 2016-05-20
Support empty description


cb372:
If you create a new role and leave the description blank, you get an error page showing a Dynamo exception. This is because we try to save a Dynamo attribute with an empty string value.

Workaround: replace empty string with `" "` before saving to Dynamo?

jharewinton:
Description is now fully optional, as per PR https://github.com/guardian/amigo/pull/112

PR #112, opened 2017-01-27
make description for recipes optional

jharewinton:
Fix for https://github.com/guardian/amigo/issues/28

DynamoDB doesn't support empty strings, so more a workaround than anything else


jharewinton:
good point @philwills I've not made it properly optional

PR #113, opened 2017-01-31
Include roles for elk stack

susiecoleman:
Written a collection of roles needed to create a recipe for an elk stack. The ami for the tools elk stack currently uses [machine images](https://github.com/guardian/machine-images/tree/master/packer/resources/features/elk-stack)

@sihil @adamnfish @TBonnin it would be great to get some feedback on this. Especially any thoughts on  what I've split into each role.

This is part of a health task on Editorial tools to update our elk stack. This is part 1 moving from machine images to amigo. The second step is moving to a more recent version of elasticsearch


roles/elasticsearch-1.7/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  file: path=/etc/apt/sources.list.d/elasticsearch.list state=touch
+
+- name: Copy repository definition to sources file
+  copy: content="deb http://packages.elastic.co/elasticsearch/1.7/debian stable main" dest=/etc/apt/sources.list.d/elasticsearch.list
+
+- name: Install elasticsearch
+  apt: name=elasticsearch state=latest
+
+- name: Install elasticsearch curator
+  pip: name=elasticsearch-curator
```
TBonnin: Should the installation of elasticsearch-curator be in the `es_housekeeping` role?

roles/elasticsearch-plugins/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Install elasticsearch plugins
+  shell: /usr/share/elasticsearch/bin/plugin --install {{ item }}
+  with_items: "{{ plugins }}"
```
TBonnin: Why not doing this in the `elasticsearch` role?

roles/elk-stack-config/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Include elasticsearch template for logstash
+  copy: src=elasticsearch-template.json dest=/etc/logstash/
+
+- name: Configure logstash input and output
+  copy: src=logstash-indexer.conf.template dest=/etc/logstash/conf.d/
```
TBonnin: I feel like the `elasticsearch template` and `logstash-indexer` configs can be quite different from one ELK stack to the other. I am not sure we should have role for it, and let whoever uses the elasticserach or logstash roles to configure it at runtime.

roles/golang/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: install golang
+  apt: name={{item}} update_cache=yes
+  with_items:
+    - golang
```
TBonnin: Should we have a default `version`, so whoever want to use this role but want a different version can overwrite it?

roles/google-auth-proxy/files/sysconfig-oauth2-proxy
```diff
@@ -0,0 +1,14 @@
...
+
+# Command line options
+OPTS="--client-id=@@CLIENT_ID \
+        --client-secret=@@CLIENT_SECRET \
+        --cookie-secret=@@COOKIE_SECRET \
+        --cookie-secure=true \
+        --email-domain=@@ALLOWED_DOMAIN \
+        --http-address=0.0.0.0:4180 \
+        --redirect-url=https://@@DOMAIN_NAME/oauth2/callback \
+        --skip-auth-regex='/__es/$' \
```
TBonnin: this line is specific to using `oauth2_proxy` within an ELK stack. Maybe we can make this another parameter?

roles/google-auth-proxy/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Download oauth proxy
+  get_url: url=https://github.com/bitly/oauth2_proxy/releases/download/v2.0/oauth2_proxy-2.0.linux-amd64.go1.4.2.tar.gz dest=/tmp
```
TBonnin: can we add an `oauth_proxy` and `go` default var, so it can be overwritten if a more recent version is required?

roles/google-auth-proxy/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- name: Unarchive
+  shell: tar -zxf /tmp/oauth2_proxy-2.0.linux-amd64.go1.4.2.tar.gz -C /opt
```
TBonnin: There is an `unarchive` module in Ansible you can use here http://docs.ansible.com/ansible/unarchive_module.html
It takes care of downloading and unzipping

roles/logstash-plugins/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Install logstash plugins
+  shell: su - logstash -s /bin/sh -c /opt/logstash/bin/plugin install {{ item }}
+  with_items: "{{ plugins }}"
```
TBonnin: Same comment as Elasticsearch: Can we add this step in the logstash roles (with empty default)

roles/kibana/files/kibana-index-backup
```diff
@@ -0,0 +1,19 @@
...
+if [ "$(curl -s 'localhost:9200/_cat/master' | awk '{print $2}')" == "$(hostname)" ]; then
+
+    # Initiate backup
+    DATE=$(date +'%Y%m%d%H%M')
+    HTTP_CODE=$(curl -s -o /tmp/kibana-backup.log -w '%{http_code}' \
+       -XPUT "localhost:9200/_snapshot/s3_backup/kibana_snapshot_${DATE}?pretty" -d '{
+         "indices": ".kibana",
+         "ignore_unavailable": "true",
+         "include_global_state": false
+    }')
```
TBonnin: I am curious: Where is the backup stored?

roles/kibana/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+
+- name: Create directory for kibana logs
+  file: path=/var/log/kibana state=directory group=kibana owner=kibana
+
+- name: Create systemd file for kibana
+  copy: src={{ systemd_file }} dest=/etc/systemd/system/kibana.service
+
+- name: Install kibana ES index backup script
+  copy: src=kibana-index-backup dest=/etc/cron.daily/kibana-index-backup mode=0755
+
```
TBonnin: I am not sure the backup step should be added to a AMI. That seems to me like it is part of configuring Kibana. Which can be quite different from one ELK stack to the other.

roles/memory-disk-autoscaling-monitoring-cloudwatch/tasks/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: Create cron job to report memory, disk space utilisation and aggregated metrics to the autoscaling group to cloudwatch
+  cron: name="log metrics to cloudwatch" user=root minute=*/5 job="/usr/local/aws-scripts-mon/mon-put-instance-data.pl --disk-space-util --mem-util --disk-path=/ --disk-path=/data --auto-scaling --from-cron"
```
TBonnin: Isn't this role already exist?

roles/nginx/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Install nginx
+  apt: name=nginx state=present
+
+- name: Install config
+  copy: src={{ config }} dest=/etc/nginx/sites-available/default
```
TBonnin: can you use the existing `nginx-extras` roles?

roles/elasticsearch-1.7/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  shell: apt-key add /tmp/GPG-KEY-elasticsearch
+
+- name: Create elasticsearch source directory
+  file: path=/etc/apt/sources.list.d state=directory
+
+- name: Create elasticsearch sources file
+  file: path=/etc/apt/sources.list.d/elasticsearch.list state=touch
+
+- name: Copy repository definition to sources file
+  copy: content="deb http://packages.elastic.co/elasticsearch/1.7/debian stable main" dest=/etc/apt/sources.list.d/elasticsearch.list
```
LATaylor-guardian: Why are we installing a very old version of ES?

susiecoleman: This is a work in 2 parts as part of a health task we are doing on tools. This is the first step, moving from machine images to amigo. Then upgrading to a newer version of elasticsearch as I realise this is really old. I would hope that we could remove this role as part of the task to update the elasticsearch version?. Updated the PR description.

roles/elasticsearch-1.7/tasks/main.yml
susiecoleman: Yes, moved

roles/elasticsearch-plugins/tasks/main.yml
susiecoleman: I got too modular. Moved this into the elasticsearch role

roles/elk-stack-config/tasks/main.yml
susiecoleman: Happy to take this role out to reduce the number of roles in amigo

roles/logstash-plugins/tasks/main.yml
susiecoleman: Fixed

roles/golang/tasks/main.yml
susiecoleman: Included a parameter for this. Although when testing this out 1.2.1 was the latest version of golang I could install on a ubuntu box. I thought there was a newer version...

susiecoleman: This was proving to be a bit of a nightmare to get working so I'm going to leave it for the moment

roles/kibana/files/kibana-index-backup
susiecoleman: s3 bucket in the composer account

roles/google-auth-proxy/tasks/main.yml
susiecoleman: Fixed

roles/google-auth-proxy/tasks/main.yml
susiecoleman: Fixed

roles/kibana/tasks/main.yml
susiecoleman: Fixed

roles/golang/tasks/main.yml
TBonnin: Would something like this be working? https://github.com/guardian/amigo/commit/c966bcef07c5486684117573889b358aa5ea8ccd

roles/logstash/tasks/main.yml
```diff
@@ -9,4 +9,8 @@
   user: name={{logstash_user}} system=yes createhome=yes
 
 - name: Create systemd file for Logstash
-  template: src=logstash.service.template dest=/etc/systemd/system/logstash.service
+  template: src={{logstash_systemd_file}} dest=/etc/systemd/system/logstash.service
+
+- name: Install logstash plugins
+  shell: su - logstash -s /bin/sh -c /opt/logstash/bin/plugin install {{ item }}
```
TBonnin: you can use 
```
become: true
become_user: "logstash"
```
rather than using `sudo`

http://docs.ansible.com/ansible/become.html

roles/memory-disk-autoscaling-monitoring-cloudwatch/tasks/main.yml
susiecoleman: This role is essentially identical to the `disk-available-cronjob` role. The issue I had was that they had slightly different parameters and I wasn't sure how best to handle parameterising this

roles/nginx/tasks/main.yml
susiecoleman: Updated nginx-extras role instead to optionally take a config file.

roles/google-auth-proxy/files/sysconfig-oauth2-proxy
susiecoleman: Done

roles/memory-disk-autoscaling-monitoring-cloudwatch/tasks/main.yml
susiecoleman: I think that would work
Also one role uses `--disk-space-util` and `--mem-util` the other uses `--disk-space-avail` `--mem-avail` any thoughts on the best way to include them?

TBonnin: What about parameters with sensible defaults
```
monitorDiskSpaceAvailable: true
monitorMemAvailable: false
...
```

And then 
```
- name: set mem_avail_option
  set_fact:
    mem_avail_option: "--mem-avail"
    when: monitorMemAvailable

cron: ... {{ mem_avail_option | default("") }} ... --from-cron
```

susiecoleman: Ended up writing a general purpose role for setting up cron jobs for cloudwatch monitoring as the other one was too specific and this new role was also too specific. Ideally people using `disk-available-cronjob` should move over to using this new role so we can deprecate the `disk-available-cronjob` role

roles/logstash/tasks/main.yml
susiecoleman: Changed to using this

roles/elasticsearch-housekeeping-cronjob/README.md
```diff
@@ -0,0 +1,2 @@
+# elasticsearch-housekeeping-cronjob
+This removes logs older than 7 days
```
TBonnin: the `7` could be parameterized I guess but not a big deal

roles/logstash/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 logstash_version: 1.5.6
 logstash_user: logstash
 logstash_config_file: /etc/logstash.conf
+logstash_systemd_file: logstash.service.template
```
TBonnin: You haven't started it, but I would not but the config_file and systemd_file vars in `defaults` because it cannot be overwritten.
Either put it in `vars/main.yml` or directly hardcoded in `tasks/main.yaml`

Also why adding `.template` extension since are copying as is into its destination

roles/nginx-extras/files/nginx-config
```diff
@@ -0,0 +1,21 @@
+server {
+    listen 80;
+
+    # Rewrite and proxy ES on __es/
```
TBonnin: This is weird to have a Elastic search specific config in the nginx role. No?

roles/logstash/defaults/main.yml
susiecoleman: The logstash role previously hardcoded the name of the file so doesn't this need to be in defaults to stop this role being broken elsewhere?

roles/nginx-extras/files/nginx-config
susiecoleman: Do you mean it's weird in the context of amigo? Or weird in the context of an nginx config.

i.e. Is the fix adding the config file in the cloudformation step or is the fix changing the file content?

TBonnin: I would say this is about configuring nginx and therefore I would do this at runtime (ie: cfn)

roles/logstash/defaults/main.yml
TBonnin: Oh I see, there is two systemd configs now. Sorry I misunderstood at first
That's quite confusing to have 2 files and I am not even sure you need all the parameters specified in `logstash_tools.service.template`
Have you seen any issues with using the original `logstash.service.template`?

roles/nginx-extras/files/nginx-config
susiecoleman: Ok will remove this and say this needs to be added in the cloudformation

roles/golang/tasks/main.yml
susiecoleman: Talked to @TBonnin this didn't solve the problem so leaving it for the moment

roles/logstash/defaults/main.yml
susiecoleman: Going to investigate in a different PR if the `logstash_tools.service.template` can go

PR #115, opened 2017-02-03
Rename role to use - instead of .

susiecoleman:
Previously the elasticsearch role was called elasticsearch-1.7 The `custom variables` field was not rendering correctly as the parsing of a directory name with a `.` in was causing issues. Renamed it to elasticsearch-1-7 as a quick fix.


philmcmahon:
👍

PR #117, opened 2017-02-13
Update deployment to use riff-raff.yaml

TBonnin:
Also add an ami-cloudformation-parameter step to upgrade AMI when amigo is deployed


jennysivapalan:
Updating Amigo with regularly up to date AMIs built in Amigo. Nice! 👍 🎆

PR #122, opened 2017-03-16
Mark unused recipes

TBonnin:
Show in the UI if a recipe is being used. (ie: one of the bake AMI for a given recipe is used by an instance or a launch configuration)

I decided to show a 3rd `unused` column in `/recipes` and a short phrase in `/recipes/RECIPE` that details how many instances and/or launch configurations are running on a recipe's AMI. Let me know what you think

screen shot 2017-03-16 at 17 12 17
screen shot 2017-03-16 at 17 12 24
screen shot 2017-03-16 at 17 12 29


app/controllers/Auth.scala
```diff
@@ -1,14 +1,12 @@
...
 import play.api.libs.json.Json
 import play.api.libs.ws.WSClient
 import play.api.mvc._
+import scala.concurrent.{ ExecutionContext, Future }
 
-import scala.concurrent.Future
-import scala.concurrent.ExecutionContext.Implicits.global
-
-class Auth(val authConfig: GoogleAuthConfig)(implicit ws: WSClient) extends Controller with AuthActions {
+class Auth(val authConfig: GoogleAuthConfig)(implicit ws: WSClient, ec: ExecutionContext) extends Controller with AuthActions {
```
TBonnin: Not related to this PR. I was only getting rid of `import scala.concurrent.ExecutionContext.Implicits.global` and injecting the `execution context` instead.

`play.api.libs.concurrent.Execution` is going to be depracated in Play 2.6 https://www.playframework.com/documentation/2.6.x/Migration26#Remove-deprecated-play.Routes

app/views/showRecipe.scala.html
```diff
@@ -18,6 +23,11 @@
       <p>Modified @fragments.timestamp(recipe.modifiedAt, recipe.modifiedBy)</p>
       <p>Build schedule: <code>@recipe.bakeSchedule.map(_.quartzCronExpression).getOrElse("(none)")</code></p>
       <p>@recipe.description</p>
+      @if(instanceCount + launchConfigurationCount > 0) {
+        <p>This recipe is used by @instanceCount instance@if(instanceCount > 1){s} and @launchConfigurationCount launch configuration@if(launchConfigurationCount > 1){s}</p>
+      } else {
+        <p><b>This recipe is not used</b></p>
```
TBonnin: I am happy to tweak the copy here

app/services/PrismAgents.scala
```diff
@@ -0,0 +1,64 @@
...
+  private val instancesAgent: Agent[Seq[Instance]] = Agent(Seq.empty)
+  private val launchConfigurationsAgent: Agent[Seq[LaunchConfiguration]] = Agent(Seq.empty)
+
+  def allInstances: Seq[Instance] = instancesAgent.get
+  def allLaunchConfigurations: Seq[LaunchConfiguration] = launchConfigurationsAgent.get
+
+  if (environment.mode != Mode.Test) {
+    refresh
+
+    val prismDataSubscription: Subscription = Observable.interval(1.minutes).subscribe { i =>
```
sihil: this could also be done with an akka scheduler and avoid bringing in the RX library

test/prism/PrismSpec.scala
```diff
@@ -1,57 +1,54 @@
...
-        |      "itemCount": 0,
-        |      "age": 38,
-        |      "stale": false
-        |    }
-        |  }]
-        |}
-      """.stripMargin
-    )
-    Prism.extractAccountNumbers(json) should be(Some(Seq("1234", "5678")))
+  def withPrismClient[T](block: Prism => T): T = {
```
thomaska: Nice refactoring

app/views/recipes.scala.html
```diff
@@ -16,12 +19,14 @@
     <thead>
       <th>Name</th>
       <th>Description</th>
+      <th>Unused?</th>
```
sihil: Could this be a used column with a little more detail (i.e. usage count)? Feels like the next question I'd have would be 'how much and where is it used?'
Just seen that it is in the detail page, but would be nice to have it surfaced here (we've done the work already).

app/controllers/RecipeController.scala
```diff
@@ -4,29 +4,41 @@ import com.gu.googleauth.GoogleAuthConfig
...
 
   def showRecipe(id: RecipeId) = AuthAction { implicit request =>
     Recipes.findById(id).fold[Result](NotFound) { recipe =>
-      val recentBakes = Bakes.list(id, limit = 20)
-      Ok(views.html.showRecipe(recipe, recentBakes))
+      val bakes = Bakes.list(recipe.id)
+      val recipeAmiIds = bakes.flatMap(_.amiId.map(_.value)).toList
+      val instancesCount = prismAgents.allInstances.count(instance => recipeAmiIds.contains(instance.imageId))
+      val launchConfigurationsCount = prismAgents.allLaunchConfigurations.count(lc => recipeAmiIds.contains(lc.imageId))
+      Ok(views.html.showRecipe(recipe, bakes.take(20), instancesCount, launchConfigurationsCount))
```
sihil: Maybe the logic here could be pulled out to make it more easily testable?

app/services/PrismAgents.scala
TBonnin: Good point. I will do that rather than importing an additional library

app/views/recipes.scala.html
TBonnin: Good point. 
I was thinking using icons but while finding a image to represent an instance is quite easy I couldn't find anything very good for launch config.

TBonnin:
@sihil you are right, it is better to flip it to show affirmative info. I am gonna think about a good UI for that. Thank you for the review

app/views/recipes.scala.html
sihil: spanner for 'config'? slightly tangential, but i think it should work.

TBonnin:
@sihil 
Requested changes have been implemented:
- Moving usage logic out of the controller
- using akka scheduler rather than importing rx.Observable
- show usage stats in recipes list. It looks like this:
screen shot 2017-03-20 at 12 18 20

Let me know what you think?

public/images/spanner.svg
```diff
@@ -0,0 +1,4 @@
+<svg fill="#000000" height="18" viewBox="0 0 24 24" width="18" xmlns="http://www.w3.org/2000/svg">
+    <path clip-rule="evenodd" d="M0 0h24v24H0z" fill="none"/>
+    <path d="M22.7 19l-9.1-9.1c.9-2.3.4-5-1.5-6.9-2-2-5-2.4-7.4-1.3L9 6 6 9 1.6 4.7C.4 7.1.9 10.1 2.9 12.1c1.9 1.9 4.6 2.4 6.9 1.5l9.1 9.1c.4.4 1 .4 1.4 0l2.3-2.3c.5-.4.5-1.1.1-1.4z"/>
+</svg>
```
sihil: I don't enormously object to adding these, providing the licensing is OK. However the project is already using bootstrap and so should have access to the glyphicons that are available in the framework. Having looked there is not a computer as far as I can see, which is sad.

app/views/recipes.scala.html
```diff
@@ -16,12 +19,21 @@
...
     <tbody>
     @for(recipe <- recipes) {
       <tr>
         <td class="has-block-link"><a href="@routes.RecipeController.showRecipe(recipe.id)" class="block-link">@recipe.id</a></td>
         <td class="has-block-link"><a href="@routes.RecipeController.showRecipe(recipe.id)" class="block-link">@recipe.description</a></td>
+        @defining(usages.getOrElse(recipe, prism.RecipeUsage.noUsage)) { usage =>
+        <td class="recipe-usage" title="Used in @usage.instances.size instance@if(usage.instances.size > 1){s} and @usage.launchConfigurations.size launch configuration@if(usage.launchConfigurations.size > 1){s}">
+            <img src="@routes.Assets.versioned("images/computer.svg")" width="16" alt="Instances" />
+            <span>@usage.instances.size</span>
+            <img src="@routes.Assets.versioned("images/spanner.svg")" width="14" alt="Launch Configurations" />
```
sihil: 👍  for alt / tooltip

app/views/showRecipe.scala.html
```diff
@@ -18,6 +22,11 @@
       <p>Modified @fragments.timestamp(recipe.modifiedAt, recipe.modifiedBy)</p>
       <p>Build schedule: <code>@recipe.bakeSchedule.map(_.quartzCronExpression).getOrElse("(none)")</code></p>
       <p>@recipe.description</p>
+      @if(usage.instances.size + usage.launchConfigurations.size > 0) {
+        <p>This recipe is used by @usage.instances.size instance@if(usage.instances.size > 1){s} and @usage.launchConfigurations.size launch configuration@if(usage.launchConfigurations.size > 1){s}</p>
```
sihil: small thing, but this should be `size != 1` as no instances would use the plural, i.e. "used by 0 instances".

TBonnin: Thanks. A rule in english I never remember

public/images/spanner.svg
TBonnin: Yep. I wanted to use the bootstrap glyphicons but as you mentionned there is no computer icon 😢

PR #125, opened 2017-03-22
Add transitive role dependencies

thomaska:
In the list of dependencies for each role, show also the transitive ones.
Add a tab with all the Roles and Recipes that use the specific Role.

Before:
For example previously the `disk-available-cronjob` role would show only the `cloud-watch-motitoring` as a dependency.
screen shot 2017-03-22 at 15 20 37

After:
It would also show `apt` and `perl-useful-packages` which are dependencies of `cloud-watch-motitoring`.
screen shot 2017-03-28 at 17 28 47

"Used by" tab:
screen shot 2017-03-28 at 17 28 57

### DIFF
--- a/app/data/Roles.scala
+++ b/app/data/Roles.scala
@@ -3,7 +3,7 @@ package data
 import java.nio.file.{ Files, Paths }
 
 import ansible.RoleParser
-import models.{ Dependency, RoleId, RoleSummary }
+import models._
 
 import scala.collection.JavaConverters._
 
@@ -34,5 +34,9 @@ object Roles {
 
   def usedBy(allRoles: Seq[RoleSummary], roleToAnalyse: RoleSummary): Seq[RoleId] = {
     allRoles.filter(_.dependsOn.contains(roleToAnalyse.roleId)).distinct.map((r: RoleSummary) => r.roleId).sortBy(_.value)
+  }
+
+  def usedByRecipes(allRecipes: Seq[Recipe], roleToAnalyse: RoleSummary): Seq[RecipeId] = {
+    allRecipes.filter(_.roles.map(_.roleId).contains(roleToAnalyse.roleId)).map(_.id)
   }
 }
--- a/app/data/Roles.scala
+++ b/app/data/Roles.scala
@@ -5,6 +5,7 @@ import java.nio.file.{ Files, Paths }
 import ansible.RoleParser
 import models.{ RoleId, RoleSummary }
 
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
 object Roles {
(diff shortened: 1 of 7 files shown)


thomaska:
[Better way to show all roles (and their dependencies) in recipe](https://trello.com/c/5MBdPgZu/23-better-way-to-show-all-roles-and-their-dependencies-in-recipe)

TBonnin:
Should the indirect dependencies displayed differently? ex: greyed out and under their direct dependencies
Ex:
```
- dep1
    - dep 10 (greyed out)
    - dep 11 (greyed out)
- dep 2
- dep 3
    - dep 30
```

Also would it make sense to also show the dependencies in the Recipe details/edit screen, so  it is easier to understand what is going to be included in the image? https://amigo.gutools.co.uk/recipes/ubuntu-xenial-java8[/edit]

thomaska:
Yeah @TBonnin makes sense to know where one dependency came from, otherwise you would be clicking through.

TBonnin:
Are the 2 latest commits changes the display of the dependencies. If yes how does that look?

thomaska:
Yes! My bad, I didn't add any screenshots!
Here's one:
screen shot 2017-03-28 at 15 20 58

thomaska:
After discussing with @TBonnin, I will move the "used by" section of the Role to different tab(s).
Showing the transitive role dependencies in the Recipe view, as @sihil suggested, will be added in a separate PR, if that's ok?!

app/views/fragments/dependencyList.scala.html
```diff
@@ -0,0 +1,11 @@
+@import models.Dependency
+@(dependencies: Set[Dependency])
+
+<ul>
+  @for(dep <- dependencies) {
+    <li>
+        <a href="#@dep.roleId">@dep.roleId</a>
+        @views.html.fragments.dependencyList(dep.dependencies)
```
philwills: Recursive templates!

app/data/Roles.scala
```diff
@@ -19,4 +19,24 @@ object Roles {
 
   def findById(id: RoleId) = list.find(_ == id)
 
+  def transitiveDependencies(allRoles: Seq[RoleSummary], roleToAnalyse: RoleSummary): Dependency = {
+    def dependencies(roleId: RoleId): Set[RoleId] = {
+      val summaries: Set[RoleSummary] = allRoles.find(r => r.roleId == roleId).toSet
```
TBonnin: nitpicking: you should be able to do `find(_.roleId == roleId)`

app/data/Roles.scala
```diff
@@ -19,4 +19,24 @@ object Roles {
...
+
+    def go(roleId: RoleId): Dependency = {
+      val children = dependencies(roleId).map(go)
+      Dependency(roleId, children)
+    }
+    go(roleToAnalyse.roleId)
+  }
+
+  def usedBy(allRoles: Seq[RoleSummary], roleToAnalyse: RoleSummary): Seq[RoleId] = {
+    allRoles.filter(_.dependsOn.contains(roleToAnalyse.roleId)).distinct.map((r: RoleSummary) => r.roleId).sortBy(_.value)
```
TBonnin: I like to break those kind of long statement into multiple lines to make it more readable but that a personal preference.
Ex:
```
allRoles
   .filter(_.dependsOn.contains(roleToAnalyse.roleId))
   .distinct
   .map(_.roleId)
   .sortBy(_.value)
```

app/data/Roles.scala
```diff
@@ -19,4 +19,24 @@ object Roles {
...
+  }
+
+  def usedByRecipes(allRecipes: Seq[Recipe], roleToAnalyse: RoleSummary): Seq[RecipeId] = {
+    allRecipes.filter(_.roles.map(_.roleId).contains(roleToAnalyse.roleId)).map(_.id)
```
TBonnin: Same comment here regarding adding new lines

app/data/Roles.scala
thomaska: No this is not nitpicking...I had the impression the compiler was complaining about this, but now that I try it, it works 😕 ...anyway I will amend!

PR #126, opened 2017-03-31
Show roles dependencies in recipes

thomaska:
Show the indirect Role dependencies in Recipes.
This is shown in the "View recipe" view, "Edit recipe" view as well as the relevant base image (view and edit) ones.

So this now looks likes this (in view recipe):
screen shot 2017-03-31 at 15 49 41

And in the Edit recipe view:
screen shot 2017-03-31 at 15 48 44
(Please note in this view if a dependency is included twice, it is shown twice.
(many thanks to @gustavpursche for his css help!)

**Update**:

After consulting a UX person (Mario) he suggested that there should be some consistency in the roles in the list and edit view so that the user doesn't get confused. 
Also he suggested adding arrows instead of bullets when depicting dependencies.
So in the spirit of those changes, all the roles are depicted using a monospace font. 
Here is how it looks now.
View recipe:
screen shot 2017-04-03 at 16 05 36

View role:
screen shot 2017-04-03 at 16 05 47

Edit recipe:
screen shot 2017-04-04 at 14 22 04

### DIFF
--- a/app/controllers/BaseImageController.scala
+++ b/app/controllers/BaseImageController.scala
@@ -28,14 +28,14 @@ class BaseImageController(
         image.amiId,
         image.linuxDist.getOrElse(Ubuntu)
       ))
-      Ok(views.html.editBaseImage(image, form, Roles.listIds))
+      Ok(views.html.editBaseImage(image, form, Roles.list))
     }
   }
 
   def updateBaseImage(id: BaseImageId) = AuthAction(BodyParsers.parse.urlFormEncoded) { implicit request =>
     BaseImages.findById(id).fold[Result](NotFound) { image =>
       Forms.editBaseImage.bindFromRequest.fold({ formWithErrors =>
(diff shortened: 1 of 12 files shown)


TBonnin:
I think we need to a better way to show those are dependencies. It is not obvious.

thomaska:
[Show all roles (and their dependencies) in the recipes](https://trello.com/c/MDZ8Y2ai/26-show-all-roles-and-their-dependencies-in-the-recipes)

thomaska:
Yes I think you are right...I 'll check with a UX person, maybe has a better idea!

TBonnin:
Code looks good to me.
I am not 100% convinced by the UI in the `/recipe/edit` page. I feel like red/rose is attracting too much the attention for an information which is secondary. Also the notion of dependencies (which is shown via nested indentation in the other screen) is missing.

johnduffell:
this is very old! :skull_and_crossbones: is it going out?

TBonnin:
I don't think there is any plan to work on this anytime soon. Closing

PR #129, opened 2017-04-10
Elasticsearch 5 and Neo4j roles

mbarton:
Merge in the work @TBonnin did to add an Elasticsearch 5 role.

Also adds a role to install the latest neo4j using apt.


roles/elasticsearch-5/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+es_major_version: 5.x
+es_version: 5.2
```
TBonnin: isn't `5.3` the latest version?

roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  file: path=/etc/apt/sources.list.d state=directory
+
+- name: Create neo4j sources file
+  file: path=/etc/apt/sources.list.d/neo4j.list state=touch
+
+- name: Copy repository definition to sources file
+  copy: content="deb https://debian.neo4j.org/repo stable/" dest=/etc/apt/sources.list.d/neo4j.list
+
+- name: Install neo4j
+  apt: name=neo4j state=latest update_cache=yes
```
TBonnin: you should be able to use ansible `apt_key` and `apt_repository` modules to achieve the same goal in fewer steps. (I have not tested it so double check the syntax)
```
- name: Install neo4j GPG signing key
  apt_key: url=https://debian.neo4j.org/neotechnology.gpg.key state=present

- name: Add neo4j repository
  apt_repository: repo='deb http://debian.neo4j.org/repo stable/' state=present

- name: Install neo4j packages
  apt: name= neo4j state=latest update_cache=yes
```

Also should we explicitly install a given neo4j version? Using `stable` would mean that running the same recipe in 6 months might install a newer version, which might create some issue in case of incompatibility.

roles/elasticsearch-5/defaults/main.yml
mbarton: Yep, I've updated it

roles/neo4j/tasks/main.yml
mbarton: Agreed on both fronts. I've update the branch accordingly

PR #132, opened 2017-04-18
Pin version of curator and bump retention time to 14 days

mbarton:
NB: I've applied the same changes on the production boxes themselves.

Fixes a problem where the cron scripts to delete indices were not running so the ES boxes ran out of disk space.

Due to the CLI for Curator changing significantly (and with [potential incompatibility problems](https://pypi.python.org/pypi/elasticsearch-curator#compatibility-matrix) as well), I've pinned us to the older version and added a note that this role only works with the older ElasticSearch.

I've also bumped up the retention period to 14 days now that we have space.

### DIFF
--- a/roles/elasticsearch-housekeeping-cronjob/files/housekeeping.sh
+++ b/roles/elasticsearch-housekeeping-cronjob/files/housekeeping.sh
@@ -10,4 +10,4 @@
 /usr/local/bin/curator --master-only delete indices --timestring '%Y.%m.%d' --time-unit days --older-than 14 2>&1 | cut -d " " -f 3- | logger -t curator
 
 # Optimise older indexes
-/usr/local/bin/curator --master-only optimize indices --timestring '%Y.%m.%d' --time-unit days --older-than 14 2>&1 | cut -d " " -f 3- | logger -t curator
+/usr/local/bin/curator --master-only optimize indices --timestring '%Y.%m.%d' --time-unit days --older-than 1 2>&1 | cut -d " " -f 3- | logger -t curator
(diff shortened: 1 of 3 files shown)


roles/elasticsearch-housekeeping-cronjob/files/housekeeping.sh
```diff
@@ -4,10 +4,10 @@
...
 # Delete old kibana index snapshots
-/usr/local/bin/curator --master-only delete snapshots --repository s3_backup --timestring '%Y%m%d%H%M' --time-unit days --older-than 7 2>&1 | cut -d " " -f 3- | logger -t curator
+/usr/local/bin/curator --master-only delete snapshots --repository s3_backup --timestring '%Y%m%d%H%M' --time-unit days --older-than 14 2>&1 | cut -d " " -f 3- | logger -t curator
 
 # Cleanup indexes
-/usr/local/bin/curator --master-only delete indices --timestring '%Y.%m.%d' --time-unit days --older-than 7 2>&1 | cut -d " " -f 3- | logger -t curator
+/usr/local/bin/curator --master-only delete indices --timestring '%Y.%m.%d' --time-unit days --older-than 14 2>&1 | cut -d " " -f 3- | logger -t curator
 
 # Optimise older indexes
-/usr/local/bin/curator --master-only optimize indices --timestring '%Y.%m.%d' --time-unit days --older-than 1 2>&1 | cut -d " " -f 3- | logger -t curator
```
sihil: This should still be 1. IIRC this optimises / freezes indexes that are no longer receiving new records.

mbarton: Oops. Absolutely right

PR #134, opened 2017-06-30
Bastion SSH role

mbarton:
I've copied across the configuration for `sshd` on a bastion instance from how it is done by the identity team.

- Modified `sshd_config` to do things like listen on a different port etc
- Remove `ubuntu` from sudoers

I didn't copy across some entries that made `/home/ubuntu/.ssh` owned by root and the `authorized_keys` file immutable, since that would conflict with the existing Amigo `s3-ssh-keys` role which automatically updates the file.

### DIFF
--- a/roles/bastion-ssh/README.md
+++ b/roles/bastion-ssh/README.md
@@ -1,0 +1,7 @@
+# Bastion SSH
+
+Additional SSH configuration for use when running as a Bastion host bridging private subnets
+with another network.
+
+It's recommended to use this role in conjunction with the `cloud-watch-logs` role pointing
+at `/var/log/auth`.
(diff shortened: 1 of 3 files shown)


philwills:
This looks reasonable to me, but my knowledge in this area is very thin.

jfsoul:
> since that would conflict with the existing Amigo s3-ssh-keys role which automatically updates the file.

Perhaps the `s3-ssh-keys` role should have some more vars to allow root ownership of the `authorized_keys` file and different permissions other than `600`.

In principle I wouldn't want us to give up security if we can change the other role.  Having said that you could decide these features aren't necessary.  Depends on your use case, what do you think?

philwills:
I think having an automated way of keeping the keys up to date is crucial, but an approach which does that in a more locked down way would be good. This feels like a reasonable step forward for now.

mbarton:
Thanks @jfsoul

PR #138, opened 2017-08-02
Skips running missed apt-daily-upgrade tasks on startup

davidfurey:
This is similar to what is disabling the apt-daily task, and should help with the issue Jacob and John were seeing: https://github.com/guardian/membership-frontend/pull/1716

### DIFF
--- a/roles/disable-missed-daily-tasks/files/apt-daily-upgrade.timer.conf
+++ b/roles/disable-missed-daily-tasks/files/apt-daily-upgrade.timer.conf
@@ -1,0 +1,2 @@
+[Timer]
+Persistent=false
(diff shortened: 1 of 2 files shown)


johnduffell:
great idea, that should certainly be a speed boost for starting instances 👍 Phil was suggesting something along those lines at lunch time too.
we need to pull this into the bake for membership before it'll take effect

davidfurey:
It is included automatically because we are using the ubuntu-xenial base image: https://amigo.gutools.co.uk/base-images/ubuntu-xenial

philmcmahon:
+💯 , thanks David!!!

philwills:
I think this sounds sane, though I don't fully understand it, but it wasn't quite what I was suggesting. When we hit this issue, it was down to the fact that `cloud-watch-logs` relies on `pip` which isn't installed by default.

https://amigo.gutools.co.uk/recipes/xenial-membership installs `pip` as an additional role, but it should probably be a dependency of `cloud-watch-logs` instead.

davidfurey:
Anyone prepared to click approve?

philwills:
Have a green tick. Also, https://github.com/guardian/amigo/pull/139

johnduffell:
oh I never "approve" anything, it seems a bit formal for me, I just give 👍 's 😆

Issue #38, opened 2016-06-02
Show dependent roles in the UI


adamnfish:
Recipes only show explicit roles in the UI. It should show dependent roles as well as those that are present by virtue of being in the provided base image.

cb372:
This information should also be shown on the edit page for a role.

e.g. if role `foo` depends on role `apt`, and you check the checkbox for `foo`, then the checkbox for `apt` should be automatically checked and greyed out.

kenoir:
This has the knock on effect of making variables in dependent roles assignable only through the parent role which is a bit confusing.

PR #144, opened 2017-11-09
Show dependencies in all the locations

sihil:
A common and fair criticism of AMIgo is that is can be hard to see what a particular recipe *actually* pulls in due to roles included in the base image and roles pulled in via dependencies.

This PR:
 - lists roles inherited from the base image when viewing recipes
 - changes the way included roles are displayed to include their dependencies on both recipe and base image pages
 - the `used by` page for a role now lists any base images that pull in a role

It is quite verbose, but I consider this better than having to view multiple pages to get the information that you need and it is now obvious when roles are implicitly pulled in.

Here's a picture of the roles section before:
screen shot 2017-11-09 at 16 55 34

and after:
screen shot 2017-11-09 at 16 47 32

### DIFF
--- a/app/views/roles.scala.html
+++ b/app/views/roles.scala.html
@@ -1,6 +1,6 @@
 @import data.Roles
 @import data.Recipes
-@(roles: Iterable[RoleSummary], recipes: Iterable[Recipe])
+@(roles: Iterable[RoleSummary], recipes: Iterable[Recipe], baseImages: Seq[BaseImage])
 @layout("AMIgo"){
 
   <h1>Roles</h1>
@@ -17,8 +17,7 @@
     </div>
 
     <div class="col-md-9">
-      <div id="sticky-anchor"></div>
-      <div class="sticky">
+      <div>
         <div id="explanation">
           Choose a role from the list to see more details.
         </div>
@@ -63,7 +62,7 @@
                 } else {
                     This role depends on (including transitive dependencies):
 
-                    @views.html.fragments.dependencyList(Roles.transitiveDependencies(roles.toSeq, role).dependencies)
+                    @views.html.fragments.dependencyList(Roles.transitiveDependencies(roles.toSeq, role.roleId).dependencies)
                 }</p>
               </div>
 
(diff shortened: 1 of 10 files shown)


app/views/showBaseImage.scala.html
```diff
@@ -29,7 +30,7 @@
   <div class="panel panel-default">
     <div class="panel-heading">Builtin roles</div>
     <div class="panel-body">
-    @fragments.customisedRoles(image.builtinRoles)
+    @fragments.customisedRoles(image.builtinRoles.map(role => role -> Roles.transitiveDependencies(allRoles, role.roleId)))
```
TBonnin: Can you not use `customisedTransitiveDependency` here? https://github.com/guardian/amigo/pull/144/files#diff-3123776b91fa002446c1ba472265a22dR35

app/views/roles.scala.html
```diff
@@ -100,6 +99,21 @@
                       }
                     }
                 </div>
+                <div class="role-usage">
+                    <h4>Base images</h4>
+                    @defining(Roles.usedByBaseImages(baseImages, role)) { usedBy =>
+                        @if(usedBy.isEmpty) {
+                            This role is not used by any base images
+                        } else {
+                            This role is included by the following base images:
```
TBonnin: included in?

cb372:
I believe this fixes the 525-day-old issue #38! 🎈

app/views/showBaseImage.scala.html
sihil: Oh yeah. That's the reason I wrote it.

app/views/roles.scala.html
sihil: I've consistently made it used by as it is everywhere else.

jharewinton:
This is awesome!

I've noticed a small issue though. When viewing roles, the base image usage doesn't take into account dependencies of the roles eg the ubuntu-init role shows that it is used in the ubuntu-xenial base image, however all the roles ubuntu-init is dependent on do not show their usage in that base image

PR #146, opened 2017-11-14
Libre Office

mbarton:
Ronseal. It's a role for the whole of Libre Office. What more could you ask for?


TBonnin:
I am curious. Where do we need to run an server with Libre Office?

mbarton:
@TBonnin I'll message you directly

PR #149, opened 2017-12-04
Support multiple teams for ssh access

nicl:
See s3-ssh-keys README change but basically, this supports providing a list of teams for the ssh prefix argument. The immediate use case if for Frontend where it is helpful to be able to re-use existing Github SSH teams (e.g. Identity-SSH-Access) rather than duplicating them.

Note, I didn't realise the full power of Jinja was available so spent a disappointing amount of time trying to coerce a Python list string representation into something usable in bash. :(

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ packer {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


philmcmahon:
👍

PR #150, opened 2017-12-05
Add chromium-browser

mbarton:
For use generating HTML previews in the investigations platform.


mbarton:
Turns out you can just use `packages` rather than a role for everything. Noice.

PR #159, opened 2018-01-09
Add user execution to GDPR scripts

currysoup:
Add missing user execution bit to GDPR log encryption roles.


currysoup:
Thanks!!!!!!

PR #161, opened 2018-01-10
add optional date format for cloudwatch log parsing

alexduf:
I'm trying to add auth.log to cloudwatch, and the date format isn't quite the same as the default play log.

This should allow us to pass whatever format we want at the time the image is booting, pushing any log, including system logs.


davidfurey:
👍

PR #162, opened 2018-01-12
Package enhancements and elasticsearch roles

sihil:
This PR adds some foundations for reducing the number of roles that people need to create by making some roles that are more generic and hopefully flexible enough to enable the installation of many more images without having to create a role.

**_Note:_** I think this is a little big but it's non-trivial to split up when I started to try. If it is a struggle to meaningfully review it, I'll spend some time doing so.

### Packages role
The `packages` roles has had the biggest overhaul. It is now possible to install packages that are not part of the Ubuntu distribution. The changes have only been made for `apt` based distributions.

First of all it is possible to install an external repository. For example you could follow the details on the [Curator DEB page](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/apt-repository.html) to install curator without creating a role:
```
packages: ['elasticsearch-curator=5.4.1'], signing_keys: ['https://packages.elastic.co/GPG-KEY-elasticsearch'], repositories: {curator: 'deb [arch=amd64] https://packages.elastic.co/curator/5/debian stable main'}
```

Secondly it is possible to install a package that is not available on the internet by hosting it in an S3 bucket. For example you could do something like:
```
s3_packages: [cerebro_0.7.2_all.deb]
```

An `ansible.packages.s3bucket` and optionally an `ansible.packages.s3prefix` must be configured in the application configuration before this can be used. The package will be downloaded and then installed using `dpkg`.

### Elasticsearch role
There is also a generic Elasticsearch role that can take `version` to specify exactly what to download and install. You can also specify plugins that will be installed. For example you might use this for a recent version with AWS support:
```
version: 6.1.1, plugins: [discovery-ec2, repository-s3]
```

### Logstash role
Similarly there is a generic Logstash role that will take a version and can optionally install plugins.
```
version: 6.1.1, plugins: [logstash-input-kinesis]
```

### Pip packages
The `pip-packages` role installs pip packages in much the same way as packages installs apt packages.

### Debug mode
When it is running somewhere other than `PROD` you will have a `bake with debug enabled` button. This runs packer with `-debug` which in turn causes packer to save the private SSH key onto the server (in the AMIgo application directory). In turn this makes it possible to SSH onto an instance that AMIgo is using to build an AMI.
screen shot 2018-01-12 at 16 35 16

### Variable support
Global ansible variables can now be supplied to a playbook which allows certain resources to be easily provided. The `s3_packages` option for the packages role uses this to provide the bucket to the role.

### Better custom role support
A few improvements were made:
 - Variable lists could just be silently dropped if one could not be parsed. It now displays an error. It's not pretty, but it is better than not noticing that the variables had been thrown away.
 - You can now specify a map/dict as the value of a variable using brackets. e.g. `variableName: {key: value}`
 - Values can now be quoted with single quotes. This is always safe to do and they will be stripped out if they are not needed. This is helpful to allow URLs as a value or strings that contain an `=` character which can be used to pin versions.


app/components/AppComponents.scala
```diff
@@ -83,21 +85,27 @@ class AppComponents(context: Context)
 
   val prism = new Prism(wsClient)
 
+  val ansibleVaribles: Map[String, String] =
```
cb372: typo 👀

sihil: I like to think of it as stylish.

app/models/CustomisedRole.scala
```diff
@@ -13,20 +13,24 @@ case class CustomisedRole(
...
+    else variables.map { case (k, v) => s"$k: ${v.quoted}" }.mkString(", ")
   }
 
 }
 
 sealed trait ParamValue { def quoted: String }
 case class SingleParamValue(param: String) extends ParamValue {
   override def toString: String = param
-  val quoted = s"'$param'"
+  val quoted = if (param.forall(CustomisedRole.allowedUnquotedChars)) s"$param" else s"'$param'"
```
mbarton: Nitpick: don't need the `s` prefix on `s"$param"`

app/models/CustomisedRole.scala
```diff
@@ -13,20 +13,24 @@ case class CustomisedRole(
...
 }
 case class ListParamValue(params: List[SingleParamValue]) extends ParamValue {
   override def toString: String = s"[${params.mkString(", ")}]"
   val quoted = s"[${params.map(_.quoted).mkString(", ")}]"
 }
+case class DictParamValue(params: Map[String, SingleParamValue]) extends ParamValue {
+  override def toString: String = s""
```
mbarton: Should this still be a useful value (for debugging etc)?

TBonnin:
@susiecoleman I think created the Elasticsearch role originally. This PR might break how ES is setup in Ed Tool

sihil:
@TBonnin Did you mean to mention me? There is currently no `elasticsearch` role - only separate "versioned" roles. Unless I misunderstand you?

TBonnin:
Sorry I meant Kibana. Is this PR change the way Kibana is being setup and run by whoever uses this role before?

sihil:
Ah yes, that's quite different. We could take the same approach as logstash and introduce a kibana-deb role?

mbarton:
I think it's only Ed Tools that is using the Kibana role so I thought it was OK to change it. Worth double checking

sihil:
Actually that is true so it would seem safe to me. Thoughts @TBonnin?

TBonnin:
Yeah I think only Ed tools is using the Kibana roles so as long as this is true I don't foresee any problems

PR #164, opened 2018-01-19
AMIgo support for encrypted bakes

sihil:
## Background

Encrypted root volumes are a good thing for GDPR but somewhat trickier to bake so requires some changes to the way we use AMIgo.

An AMI with an encrypted root volume needs to live in the account you want to launch it from (until now most AMIs have lived in the AMIgo account and launched from other accounts).

The proposed mechanism is to bake an AMI in the deploy_tools account and then accounts interested in running the AMI can copy the AMI into their own account, encrypting it in the process. This copy can be kicked off inside a Lambda in the target account that is subscribed to a notification topic in the AMIgo account.

## This PR

In order to facilitate the above this work:
 - adds the `snapshot_users` parameter to the packer build (this gives permission for other accounts to make copies of the AMI) - this requires Packer 0.12.1 or later
 - creates an SNS topic i it doesn't exist
 - waits for `AmiCreated` events and sends a message on the SNS topic when they occur
 - adds a new field to a recipe that is a comma separated list of AWS account numbers - this is sent in the message to the Lambdas and provides a simple mechanism for working out if a copy should be made for a given AMI
 - creates an `imageCopier` lambda that subscribes to the SNS topic, kicks off an image copy and applies the correct tags
 - provides a CFN template that makes it easy to setup the `imageCopier` in a target 

## Further work

Before merging
 - [x] Apply the CFN update to PROD (to allow creation of SNS topic)
 - [x] Modify Riff-Raff to find AMI images that are visible to the stack being deployed to (i.e. private images in another AWS account are not relevant, but shared images are) -
 See https://github.com/guardian/riff-raff/pull/479
 - [x] Merge and deploy Riff-Raff PR (without this other deploys might pick up AMIs they can't use)
 - [x] Verify the permissions, if any, required to launch an instance using default / user created KMS keys

Longer term
 - Consider how the lambdas should be deployed / updated
 - Improve the UI around selecting AWS accounts
 - Consider how monitoring and reporting should evolve... should AMIgo be aware of copied AMIs?


mchv:
Thanks Simon for looking at this. @tomforbes this looks interesting for our elasticsearch cluster

sihil:
You're welcome. I'm on holiday next week so if someone wants to pick this up and take it across the line then I'd be glad to hand over. Or a review of these PRs. @mchv @adamnfish @tomrf1 @alexduf

tomrf1:
👍 looks good

PR #165, opened 2018-01-26
Enable TCP Forwarding for bastion-ssh role

philmcmahon:
I want to be able to set up an SSH tunnel to instances in the discussion private vpc. Currently the only way I can SSH into an instance in the private VPC is to first SSH into the bastion, then to ssh again into the instance I want. This is sad, not least because it means I have to copy the SSH key of the target instance to the bastion instance in order for it to work. The other sad thing is that it's much slower than simply doing 'ssh ubuntu@destinationinstanceip' and having ssh config pick this up and automatically create an ssh tunnel.

With this enabled, devs can add some config like this to ~/.ssh/config and transparently use the bastion host.

```
Host 10.222.111.* 10.222.112.* 10.222.113.* 10.222.114.*
    StrictHostKeyChecking no
    UserKnownHostsFile=/dev/null
    IdentityFile ~/.ssh/mykey.pem
    ProxyCommand ssh -W %h:%p ubuntu@bastionhost-ec2.com
```

BUT this is something of a security thingy. I'd argue that since even without it anyone with the keys in the network can do what they want it's not a big change to modify this property, but others might have different thoughts...

### DIFF
--- a/roles/bastion-ssh/files/sshd_config
+++ b/roles/bastion-ssh/files/sshd_config
@@ -57,7 +57,7 @@ PasswordAuthentication no
 
 # Disable X11 and port forwarding
 X11Forwarding no
-AllowTcpForwarding no
+AllowTcpForwarding yes
 
 # Prevent the use of insecure home directory and key file permissions
 StrictModes yes


jfsoul:
> This is sad, not least because it means I have to copy the SSH key of the target instance to the bastion instance in order for it to work.

Is this the case? Surely with `ssh -A ...` you're not copying keys anywhere?

philmcmahon:
You're right, I've not used `ssh -A` before.

philmcmahon:
Maybe @aug24 might have an opinion on this?

philmcmahon:
Closing this as it's a bad thing to do - apparently switching it off is sometimes required for compliance reasons. I'm going to script around it instead.

PR #167, opened 2018-01-30
Await aws tags

mbarton:
We have seen some issues in PROD on Ed Tools caused by tags not being available when the user data script is run.

The real fix is to change the apps to not call Describe Tags to find out Stack, App, Stage etc. In the meantime we can use this role that will block calling the main script until tags are available (or 1 minute has elapsed).

### DIFF
--- a/roles/aws-await-tags/README.md
+++ b/roles/aws-await-tags/README.md
@@ -7,7 +7,7 @@ We use EC2 [Tagging](https://docs.aws.amazon.com/cli/latest/reference/ec2/descri
 to look up the `Stack`, `App` and `Stage` for a given instance.
 
 Unfortunately, the API is eventually consistent and can return empty tags on some AWS
-instance types that boot fast enough. This can cause major problems will apps that are
+instance types that boot fast enough. This can cause major problems with apps that are
 written to assume a lack of tags means run in `DEV` mode.
 
 The correct solution to the problem is to change the application to receive the information
@@ -17,4 +17,6 @@ This role exists so we can get the safety of consistently reading tags before we
 a chance to change the code in all the projects.
 
 It simply adds a script to Cloud Init [Scripts Per Boot](http://cloudinit.readthedocs.io/en/latest/topics/modules.html#scripts-per-boot)
-that tries to get the tags in a loop, failing if none are available after 1 minute.
+that tries to get the tags in a loop.
+
+NB: if tags are not found the `user-data` script **will still run**.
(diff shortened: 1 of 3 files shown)


roles/aws-await-tags/templates/await_tags.sh
```diff
@@ -0,0 +1,18 @@
...
+
+    if (( $NUM_TAGS > 0 )); then
+        echo "AWS tags loaded"
+        exit 0
+    fi
+
+    sleep 5
+done
+
+echo "!! AWS Tags not found after 1 minute !!"
```
RichieAHB: Do you need to `exit 1` here?

roles/aws-await-tags/README.md
```diff
@@ -0,0 +1,20 @@
+Await Tags
+==========
+
+This role should ideally not be used for new projects.
+
+We use EC2 [Tagging](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-tags.html)
+to look up the `Stack`, `App` and `Stage` for a given instance.
+
+Unfortunately, the API is eventually consistent and can return empty tags on some AWS
+instance types that boot fast enough. This can cause major problems will apps that are
```
RichieAHB: Also, typo `will` > `with` (to show that I'm paying attention) ...

roles/aws-await-tags/templates/await_tags.sh
RichieAHB: Also do you need to single quote anything with a `!` ... I know you need to on a mac otherwise bash will try a history substitution.

mbarton: I did have `exit 1` but cloud init was ignoring the failure and pressing ahead anyway so I removed it to avoid confusion

RichieAHB: Maybe a comment? Just because I think that'd be the first fix I'd attempt if I was going to try to get that to fail the deploy ...

mbarton: Done

sihil:
Hello from Austria!

I believe you should be able to execute this as a one time systemd unit that is required to run before cloud init. Doing so would mean it could be baked in and run without user data changes.

mbarton:
Yeah that's probably less brittle than this, my systemd knowledge isn't up to it I'm afraid.

This fix still doesn't require user data to be changed since it hooks it to the cloud init `scripts-per-boot` module that runs before `scripts-user`.

mbarton:
Also my impression of you on the mountain: 
8569b8645c1ccd99e562715637294f2c--epic-beard-the-bear

PR #168, opened 2018-02-02
Update filebeat

mbarton:
The old role installed a big mash of filebeat, logstash and kinesis and is no longer used. I've replaced it with a nice slimline install of just filebeat.


philmcmahon:
👍 ✂️

PR #169, opened 2018-02-05
Add open JDK repository on Trusty (14.04)

shtukas:
This change adds an action to role `java8`, that only affect Trusty (14.04), by adding a missing repository preventing the role to fail when running recipes on that image.


shtukas:
Haven't yet been tested on CODE

shtukas:
@sihil 
Change as per discussion with @adamnfish

shtukas:
@sihil @adamnfish 
Successfully tested on CODE.

PR #172, opened 2018-02-08
New stack to handle KMS keys for Image Copier

ajwl:
The image copier lambda requires a KMS key unique to each account it runs in.

The best way to do this is to create an additional stack that only creates KMS keys for the Image copier and then deploy this second stack in all accounts that use the image copier lambda. 

The cloudformation added here describes and creates that second stack.

### DIFF
--- a/imageCopier/imageCopier.yaml
+++ b/imageCopier/imageCopier.yaml
@@ -5,9 +5,6 @@ Parameters:
   AmigoTopicArn:
     Description: The SNS topic to subscribe to
     Type: String
-  KmsKeyARN:
-    Description: ARN of KMS key to encrypt the AMI copies
-    Type: String
   EncryptedTagValue:
     Description: The value of the Encrypted tag on the created AMI
     Type: String
@@ -71,7 +68,7 @@ Resources:
           - kms:CreateGrant
           - kms:GenerateDataKey*
           - kms:DescribeKey
-          Resource: !Ref KmsKeyARN
+          Resource: !ImportValue amigo-imagecopier-key
 
   ImageCopyPolicy:
     Type: AWS::IAM::Policy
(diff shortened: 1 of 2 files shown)


ajwl:
@sihil pr here

sihil:
I do wonder if we should consider setting a key rotation policy. I'm not sure what the default is. Any thoughts @adamnfish ?

adamnfish:
I would recommend enabling key rotation and letting Amazon use the default policy (which is currently to rotate every 365 days). It's supposed to be transparent to users, it keeps old keys around to read old data (although there are (minor) costs involved). This would achieve "compliance box ticking" with minimal effort.

adamnfish:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html

(add this to the Properties)

    EnableKeyRotation: Boolean

sihil:
@ajwl Looks like @adamnfish copied that from the CFN docs with the type rather than a value. Should be `EnableKeyRotation: true`.

sihil:
We also need to check we're able to update the stack set with the new CFN. Hopefully it won't create a new key (which would presumably be impossible?)

adamnfish:
I'm pretty sure it won't be a problem. This is an in-place settings change rather than something that requires the creation of a new resource. As you say @sihil, I pasted the signature, not the value. Sorry it wasn't clear!

ajwl:
Sorry 🙈 - fixed

sihil:
Great - can you apply the change to the stackset? Once we know that's successful I'm happy for this to be merged.

ajwl:
@sihil damn do you know what, I can't because I revoked my root access.

adamnfish:
@ajwl easy to get it back temporarily :-)

ajwl:
@sihil @adamnfish sure I'll re-pr janus

sihil:
I've applied this and confirmed working. I also added a description as it was not possible to identify the purpose of the key.

PR #174, opened 2018-02-09
Make encrypted copies visible in AMIgo

sihil:
It's pretty confusing that the usage and bake history don't currently display copies of AMIs that are in other accounts. This fixes that by adding a little more magic to both of them.

For example the bake history now looks like this:
screen shot 2018-02-09 at 18 36 45

As an added bonus I fixed the copy to clipboard functionality so that it works!

### So what changes have I made?
In short I've added a new prism agent that queries prism for all AMIs that have the `CopiedFromAMI` tag set. That is the tag that is set with the original AMI ID when the lambda kicks off a copy. This is then correlated with the list of AMIgo AMI IDs to create the list of copies.

In order to display human friendly accounts I've also expanded the data gathered from prism for accounts to include the name as well as number. This can then be used as a lookup table.

### DIFF
--- a/app/prism/Prism.scala
+++ b/app/prism/Prism.scala
@@ -9,9 +9,8 @@ import scala.concurrent.{ ExecutionContext, Future }
 class Prism(ws: WSClient, baseUrl: String = "http://prism.gutools.co.uk")(implicit ec: ExecutionContext) {
   import Prism._
 
-  def findAllAWSAccountNumbers(): Future[Seq[String]] = {
-    findAll[SourceInstance]("/sources?resource=instance&origin.vendor=aws")
-      .map(_.map(_.accountNumber))
+  def findAllAWSAccounts(): Future[Seq[AWSAccount]] = {
+    findAll[AWSAccount]("/sources?resource=instance&origin.vendor=aws")
   }
 
   def findAllInstances(): Future[Seq[Instance]] = findAll[Instance]("/instances")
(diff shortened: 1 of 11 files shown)


app/prism/Prism.scala
```diff
@@ -9,15 +9,16 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-      .map(_.map(_.accountNumber))
+  def findAllAWSAccounts(): Future[Seq[AWSAccount]] = {
+    findAll[AWSAccount]("/sources?resource=instance&origin.vendor=aws")
   }
 
   def findAllInstances(): Future[Seq[Instance]] = findAll[Instance]("/instances")
 
   def findAllLaunchConfigurations(): Future[Seq[LaunchConfiguration]] = findAll[LaunchConfiguration]("/launch-configurations")
 
+  def findCopiedImages(): Future[Seq[Image]] = findAll[Image]("/images?tags.CopiedFromAMI!=")
```
tomrf1: this query parameter looks strange

app/prism/RecipeUsage.scala
```diff
@@ -11,7 +11,12 @@ object RecipeUsage {
   def noUsage(): RecipeUsage = RecipeUsage(Seq.empty[Instance], Seq.empty[LaunchConfiguration])
 
   def apply(recipe: Recipe, bakes: Iterable[Bake])(implicit prismAgents: PrismAgents): RecipeUsage = {
-    val amiIds = bakes.flatMap(_.amiId.map(_.value)).toList
+    println(s"Finding usages for $recipe / $bakes")
```
tomrf1: `println`s intentionally left in?

app/prism/Prism.scala
sihil: True story. It's the query language I built for prism. You can filter on any field and if the query param name ends in ! then it inverts it. What I want to say here is that there is a CopiedFromAMI tag. This is really saying that the CopiedFromAMI tag isn't an empty string, but will only match when it exists at all so it's good enough for me.

You can experiment at http://prism.gutools.co.uk/images?tags.CopiedFromAMI!=

tomrf1:
👍 
excellent 💯

PR #178, opened 2018-02-22
Add Docker role

nicl:
Adds Docker role.


roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Add repo key
+  apt_key:
+    url: "https://download.docker.com/linux/debian/gpg"
+    state: present
+
+- name: Add docker.io repo
+  apt_repository: repo='deb [arch=amd64] https://download.docker.com/linux/debian xenial stable' state=present
+
+- name: Install docker
+  apt: name=docker-ce update_cache=yes state=installed force=yes
```
sihil: I really dislike floating versions. Or perhaps: I really dislike debugging problems that end up being caused by an upgrade of something you depend on downstream being implicitly pulled in.

I'd like them more (but still not a whole lot) if we had serverspec to run the finished AMI through its paces before allowing anything else to use it. For now though we should force people to chose a version that we know works.

sihil:
You might want to look at the `packages` role to see if you can do what you want to without a new role. Having said that, docker seems useful enough that a role makes sense.

franziskas:
@nicl @sihil anything to salvage here?

AWare:
Installing docker as a package in #194 doesn't leave the system with a usable docker. 

My instinct is that the docker uninstall in https://github.com/guardian/deploy-tools-platform/blob/master/teamcity/install-docker.sh is required. I'd like to add this to the Docker role and merge it into my PR. Would also parameterise the version while I'm at it. 

@sihil @franziskas @nicl all okay with this?

AWare:
@nicl I've rebased this to test on code.

AWare:
Tested.

PR #179, opened 2018-02-27
Sshd security

aug24:
Two new roles:

Allow only ubuntu to ssh on to the box.  This is intended for teams who have adopted the ssm-scala approach to temporary ssh credentials.

Do not allow root to ssh on to the box.  This is a patch for a security risk and should be adopted everywhere.

### DIFF
--- a/roles/ssh-deny-root/README.md
+++ b/roles/ssh-deny-root/README.md
@@ -1,0 +1,4 @@
+SSH Deny root
+=============
+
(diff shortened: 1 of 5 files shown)


sihil:
Can I suggest that `ssh-deny-root` becomes a dependency of `ubuntu-init`? i.e. I agree that it should be set everywhere and I think we should just adopt it.

PR #180, opened 2018-03-12
Overwrite temp keys file instead of append

jfsoul:
We're seeing our `authorized_keys` file contain many hundreds of repeated keys because it looks like we are appending instead of overwriting to /tmp/authorized_keys.

~Requires testing.~ Tested.


adamnfish:
I guess the downside is that this will lock you out if the fetch fails. Not a big deal though, and that's definitely better than "you are never removed"

philmcmahon:
Looks good to me, I think having some kind of fallback option isn't worth it now that everything should be moving towards ssm anyway

PR #181, opened 2018-03-20
Fix s3-ssh-keys role for when more than one team is declared

rtyley:
PR #180 introduced a new bug for the case when more than one team prefix is declared - the docs for `s3-ssh-keys` say this is supported:

```
ssh_keys_prefix: [Team-NameA, Team-NameB]
```

...but when overwrite (`>`) is used rather than append (`>>`) for concatenating together the keys of multiple teams this means that only the last team wins - the keys from all other teams are overwritten.

Instead in this commit we use a unique temporary aggregating directory, to allow us to join various key files together, without accumulating unwanted previous versions of the key data.

cc @amyhughes

### DIFF
--- a/roles/s3-ssh-keys/templates/install-keys.sh.template
+++ b/roles/s3-ssh-keys/templates/install-keys.sh.template
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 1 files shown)


amyhughes:
We've tested this and it seems to work, so I'm going to merge this now.

PR #184, opened 2018-03-22
Add friendly-hostnames role for giving servers human readable names

amyhughes:
This is a second attempt at PR #182 

This reverts commit 850341b778762741fbf06fb6b830979a59658571.


amyhughes:
I've tested this and have managed to give an instance a hostname using the generated AMI

PR #189, opened 2018-05-23
Exit immediately if we are unable to call describe-tags

mbarton:
Facia uses the base role that has await tags in it but does not have permission in the CloudFormation to call describe-tags in the API. It doesn't need it since it injects the stack, app and stage through the launch configuration. Unfortunately, the script still waits a minute even though each call fails because it doesn't have permission.

This PR changes the script to exit early if we don't have permission. If we have permission but the tags are not yet present we get back an empty list so it still waits as expected.

### DIFF
--- a/roles/aws-await-tags/templates/await_tags.sh
+++ b/roles/aws-await-tags/templates/await_tags.sh
@@ -7,7 +7,7 @@ for i in {1..12}
 do
     NUM_TAGS=$(aws ec2 describe-tags --filters Name=resource-type,Values=instance Name=resource-id,Values=${INSTANCE_ID} --region eu-west-1 | jq ".Tags | length")
 
-    if [[ -z $NUM_TAGS ]]; then
+    if [[ -z "$NUM_TAGS" ]]; then
         echo "Unable to call describe tags. Exiting immediately"
         exit 255
     fi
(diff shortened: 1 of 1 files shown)


roles/aws-await-tags/templates/await_tags.sh
```diff
@@ -7,6 +7,11 @@ for i in {1..12}
 do
     NUM_TAGS=$(aws ec2 describe-tags --filters Name=resource-type,Values=instance Name=resource-id,Values=${INSTANCE_ID} --region eu-west-1 | jq ".Tags | length")
 
+    if [[ -z $NUM_TAGS ]]; then
```
sihil: In general it's safer to do `if [[ -z "${NUM_TAGS}" ]]; then`. The quotes can help to avoid a zero length `NUM_TAGS` wind up evaluating as `[[ -z ]]` which makes no sense. Although in this case I presume it works? TBH I can't event remember when I started doing this, but it was certainly "a good thing".

mbarton: It does work yeah but I agree, I'll update the code. I was originally trying to test the return code but that didn't seem to work inside the `$(` even though I could see it if I ran the command manually.

BASH

PR #193, opened 2018-07-06
Add recipe deletion capabilities to AMIgo

sihil:
It's now possible to delete recipes from AMIgo!!!

When a recipe is deleted we also delete all of the AMIs and the related snapshots asynchronously. This needs to be done across all accounts so a new Lambda has been created to handle this.

Messages are fired out to a new housekeeping topic that is distributed to lambdas in all accounts much like they are when a copy is required. This is a fire and forget, we don't check they complete.

We also introduce housekeeping jobs. When a recipe is deleted we mark all of the bakes as deleted and delete the recipe and the deletion housekeeping job actually does the work of sending the AMI delete messages and cleaning up bake log and bake entries in Dynamo.

It is not possible to delete a recipe if any of the bakes are still being used by an instance or launch configuration.

This is another step towards deleting old and unused bakes as a regular housekeeping job.

Todo
-----

 - [x] Figure out how to also delete the bake logs (this is non-trivial, apparently we need to know the range key, sad times)
 - [ ] Roll out the new imageCopier CFN that includes the housekeeping Lambda via the Stackset
 - [x] Merge and deploy to PROD

### DIFF
--- a/app/AppLoader.scala
+++ b/app/AppLoader.scala
@@ -12,10 +12,10 @@ class AppLoader extends ApplicationLoader {
     val components = new AppComponents(context)
 
     Logger.info("Starting the scheduler")
-    components.bakeScheduler.start()
+    components.scheduler.start()
     components.applicationLifecycle.addStopHook { () =>
       println("Shutting down scheduler")
-      Future.successful(components.bakeScheduler.shutdown())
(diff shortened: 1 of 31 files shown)


mchv:
Nice work 💯

PR #194, opened 2018-07-10
Teamcity agent as ansible roles

AWare:
This adds some ansible roles that should define a teamcity agent, hopefully we'll be able to bake and rotate these more frequently now.


AWare:
This depends on https://github.com/guardian/amigo/pull/178, after which I'll add docker compose. 

With NVM, it requires the user to exist before installing, which I'm not sure how best to deal with.

AWare:
This builds a build agent,  I'd like to chat about how we want to put this into production.

franziskas:
@AWare can this be merged?

AWare:
Not yet.

AWare:
This has been doing a fair few teamcity builds this afternoon, so it's time to get this shipped.

AWare:
This doesn't install zip on the instance.

PR #196, opened 2018-07-12
Automatically delete any bake that is older than a month and is not in use

katebee:
### 👉Remove unused recipe argument from RecipeUsage apply

This was not being used anyway so is a bit of a code clean up

### 👉Mark old unused bakes for deletion with a housekeeping job

This adds a new housekeeping task that runs each hour and will find bakes over **30 days old** that are **not in use**. To prevent hitting DynamoDB too hard and running into throughput exceptions, only the first 100 of the old unused bakes will be marked for deletion upon each run.

We have never deleted bakes before, so it may take us a while to get under 100! Once we do, it is unlikely we will get over 100 again, so scheduling this task for once an hour and taking 100 seemed like a reasonable approach that allows us to avoid scaling up the DynamoDB throughput, or adding in a `Thread.sleep(...)` that probably wont matter after the first run...

This change means that any bakes marked as deleted will be deleted by the other housekeeping tasks added in #193. Actually deleting the bakes will be rate-limited by the BakeDeletion housekeeping task.

### TODO:

- [x] add tests for getOldUnusedBakes
- [x] test on CODE
- [ ] Remember to check the PROD logs so we can see how many bakes we end up deleting 😁

### DIFF
--- a/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
+++ b/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
@@ -9,6 +9,9 @@ import prism.RecipeUsage
 import services.PrismAgents
 
 object MarkOldUnusedBakesForDeletion {
+  val MAX_AGE = 30
+  val BATCH_SIZE = 100
+
   def getOldUnusedBakes(
     recipeIds: Set[RecipeId], now: DateTime,
     listBakes: RecipeId => Iterable[Bake],
@@ -17,7 +20,7 @@ object MarkOldUnusedBakesForDeletion {
 
     val oldBakes = allBakes.filter { bake =>
       val duration = new Duration(bake.startedAt, now)
-      duration.getStandardDays > 30
+      duration.getStandardDays > MAX_AGE
     }
     val recipeUsage = getRecipeUsage(oldBakes)
     val usedBakes = recipeUsage.bakeUsage.map(_.bake).distinct.toSet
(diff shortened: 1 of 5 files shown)


app/housekeeping/MarkOldUnusedBakesForDeletion.scala
```diff
@@ -0,0 +1,49 @@
...
+object MarkOldUnusedBakesForDeletion {
+  def getOldUnusedBakes(
+    recipeIds: Set[RecipeId], now: DateTime,
+    listBakes: RecipeId => Iterable[Bake],
+    getRecipeUsage: Iterable[Bake] => RecipeUsage): Set[Bake] = {
+    val allBakes = recipeIds.flatMap(listBakes)
+
+    val oldBakes = allBakes.filter { bake =>
+      val duration = new Duration(bake.startedAt, now)
+      duration.getStandardDays > 30
```
sihil: Pull this out into a constant or even config?

app/housekeeping/MarkOldUnusedBakesForDeletion.scala
```diff
@@ -0,0 +1,49 @@
...
+  override def housekeep(): Unit = {
+    implicit val implicitPrismAgents: PrismAgents = prismAgents
+    implicit val implicitDynamo: Dynamo = dynamo
+    Logger.info(s"Started marking old, unused bakes for deletion")
+    val now = new DateTime()
+    val recipeIds = Recipes.list().map(_.id).toSet
+    val oldUnusedBakes = MarkOldUnusedBakesForDeletion.getOldUnusedBakes(recipeIds, now, Bakes.list, RecipeUsage.apply)
+    Logger.info(s"Found ${oldUnusedBakes.size} unused bakes over 30 days old")
+
+    val bakesToMark = oldUnusedBakes.take(100)
```
sihil: As above: pull this out into a constant / config?

test/housekeeping/MarkOldUnusedBakesForDeletionSpec.scala
```diff
@@ -0,0 +1,61 @@
+package housekeeping
+
+import models._
+import org.joda.time.{ DateTime, DateTimeZone }
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{ FlatSpec, Matchers }
+import prism.{ BakeUsage, RecipeUsage }
+
+class MarkOldUnusedBakesForDeletionSpec extends FlatSpec with Matchers with MockitoSugar {
```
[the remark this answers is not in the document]
katebee: Good catch! Not any more 😁

PR #199, opened 2018-07-16
Deploy via deb

sihil:
This PR moves amigo to deploy using DEB. At the same time I've changed logging when in production to log to `/var/log/amigo/amigo.log`.

This modernises AMIgo and also sets us up for being able to ship logs from AMIgo to Central ELK (separating amigo logs from syslog).


build.sbt
```diff
@@ -1,8 +1,40 @@
...
+
+lazy val root = (project in file("."))
+  .enablePlugins(PlayScala, RiffRaffArtifact, JDebPackaging)
+  .settings(
+    packageName in Universal := normalizedName.value,
+    maintainer := "Guardian Developer Experience <devx@theguardian.com>",
+
+    serverLoading in Debian := Systemd,
+    riffRaffPackageType := (packageBin in Debian).value,
+    riffRaffBuildIdentifier := sys.env.getOrElse("TRAVIS_BUILD_NUMBER", "DEV"),
```
alexduf: This _should_ already be supported by sbt-riffraff-artifact:
https://github.com/guardian/sbt-riffraff-artifact/blob/master/src/main/scala/com/gu/riffraff/artifact/BuildInfo.scala#L46

Though I realise you're just moving the code

cloudformation.yaml
```diff
@@ -227,28 +227,13 @@ Resources:
...
-
-          mv /home/amigo/amigo-1.0-SNAPSHOT /home/amigo/amigo
-
-          mkdir /home/amigo/logs
-
-          chown -R amigo /home/amigo
-
+          aws --region eu-west-1 s3 cp s3://deploy-tools-dist/deploy/${Stage}/amigo/amigo_1.0-SNAPSHOT_all.deb /tmp/amigo.deb
+          dpkg -i /tmp/amigo.deb
           systemctl start amigo
```
alexduf: I'm pretty sure you don't need to start it, `dpk` will do it for you (it's nice like that)

sihil: Yeah - I thought that as I did that. Will remove.

build.sbt
sihil: We should get this merged and ship it. Then I'd be more confident about it working properly in all scenarios: https://github.com/guardian/sbt-riffraff-artifact/pull/43

cloudformation.yaml
AWare: oh I missed this. 

I wouldn't call that _nice_ as it makes it less easy to load any config that's readable by the user created by dpkg before starting the service.

PR #202, opened 2018-07-18
Change the name of the S3 key to force reload

sihil:
A bit of a hack that allowed us to force out a new version of the lambda to all the stacks in the stack set (by virtue of changing the s3 key).

There is another PR (https://github.com/guardian/amigo/pull/201) that should solve this properly but for the time being this let us fix a bug.

Note that the proposed riff-raff.yaml doesn't deal with dependencies quite as you expect as each stack/region combination does its own dependency resolution and Riff-Raff doesn't support dependencies across stacks.


jacobwinch:
Minor point: we've changed the S3 key not the lambda name (the PR title and description don't currently agree!)

PR #203, opened 2018-07-18
Automatically deploy AMIgo cloudformation at deploy time

sihil:
The world is much better when your deploy matches up with the cloudformation. Since the CFN template for AMIgo is in the codebase we should definitely deploy it automatically.


sihil:
This also works.

PR #204, opened 2018-07-18
Run packer with /tmp as the working directory so it can write to it

sihil:
When in debug mode packer needs to run in a directory it can write to in order to output the private key for the instance it is building. This should do the trick. Hopefully it won't break anything else 🤞


sihil:
Have tested this - works as expected.

PR #208, opened 2018-07-24
Add housekeeping task for orphaned bakes and a housekeeping UI

katebee:
Orphaned bakes may exist for a few reasons... 

If a recipe has been deleted from a table but not the associated bake, then these lonely bakes will linger on and never get picked up by the housekeeping... Until now

#### 👉 Adds a housekeeping task that hunts for orphaned bakes

Something interesting was happening with the DEV DynamoDB table and there were read errors so I have decided to surface these so the housekeeping task can decide what to do.

I have set the `FAULT_TOLERANCE` to `0` for now, which means that the housekeeping scheduler will abort in DEV mode:

picture 155

If the housekeeping task does not get stopped it should happily mark all the orphaned bakes for deletion. Once marked the deletion housekeeping should finish off doing the clean up.

#### 👉 New Housekeeping UI

This includes a form that manually allows you to mark the orphaned bakes for deletion. This is helpful if your database has read errors yet you want to delete things anyway!

picture 154

The button works and will `POST` a request that should mark all the listed bakes for deletion, irrespective of the `FAULT_TOLERANCE`. The UI needs some work and the button just redirects you to the same page therefore currently it looks like nothing happens... The application is doing something secretly I promise!

#### 👉 Bringing some of the Death by Switches spirit to AMIgo

We should not let frontend have all the fun~

#### 🦄 TODO: 

- [x] Test on DEV
- [x] Test on CODE

- [ ] Improve the UI and UX

### DIFF
--- a/app/controllers/HousekeepingController.scala
+++ b/app/controllers/HousekeepingController.scala
@@ -8,11 +8,11 @@ import play.api.Logger
 import play.api.mvc._
 
 class HousekeepingController(val authConfig: GoogleAuthConfig)(implicit dynamo: Dynamo) extends Controller with AuthActions {
-  private val (errors, recipes) = Recipes.recipesWithErrors
-  private val recipeIds = recipes.map(recipe => recipe.id).toSet
-  private val orphanedBakes = MarkOrphanedBakesForDeletion.findOrphanedBakeIds(recipeIds, Bakes.scanForAll())
 
   def showOrphans = AuthAction {
+    val (errors, recipes) = Recipes.recipesWithErrors
+    val recipeIds = recipes.map(recipe => recipe.id).toSet
+    val orphanedBakes = MarkOrphanedBakesForDeletion.findOrphanedBakeIds(recipeIds, Bakes.scanForAll())
     Ok(views.html.housekeeping(orphanedBakes, errors.length))
   }
 
--- a/app/controllers/HousekeepingController.scala
+++ b/app/controllers/HousekeepingController.scala
@@ -1,0 +1,30 @@
+package controllers
+
+import com.gu.googleauth.GoogleAuthConfig
+import data._
+import housekeeping.MarkOrphanedBakesForDeletion
+import models.BakeId
+import play.api.Logger
+import play.api.mvc._
+
+class HousekeepingController(val authConfig: GoogleAuthConfig)(implicit dynamo: Dynamo) extends Controller with AuthActions {
+  private val (errors, recipes) = Recipes.listWithErrors()
+  private val recipeIds = recipes.map(recipe => recipe.id).toSet
+  private val orphanedBakes = MarkOrphanedBakesForDeletion.findOrphanedBakeIds(recipeIds, Bakes.scanForAll())
+
+  def showOrphans = AuthAction {
+    Ok(views.html.housekeeping(orphanedBakes, errors.length))
+  }
+
+  def deleteOrphans(): Action[AnyContent] = AuthAction { implicit request =>
+    val formData = request.body.asFormUrlEncoded.get("orphaned-bake")
+
+    formData.foreach { bake =>
+      BakeId.fromString(bake) match {
+        case Right(bakeId) => Bakes.markToDelete(bakeId)
+        case Left(err) => Logger.warn(err.toString)
+      }
+    }
+    Redirect(routes.HousekeepingController.showOrphans())
+  }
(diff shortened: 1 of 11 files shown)


app/housekeeping/MarkOrphanedBakesForDeletion.scala
```diff
@@ -0,0 +1,40 @@
+package housekeeping
+
+import data.{ Bakes, Dynamo, Recipes }
+import models.{ Bake, BakeId, RecipeId }
+import org.quartz.SimpleScheduleBuilder
+import play.api.Logger
+import services.PrismAgents
+
```
jacobwinch: Might be worth adding a (brief) comment which explains what an orphaned bake is

app/housekeeping/MarkOrphanedBakesForDeletion.scala
```diff
@@ -0,0 +1,40 @@
...
+    implicit val implicitDynamo: Dynamo = dynamo
+    val (errors, recipes) = Recipes.listWithErrors
+    errors match {
+      case _ if errors.length > MarkOrphanedBakesForDeletion.FAULT_TOLERANCE =>
+        Logger.info(s"Housekeeping found ${errors.length} database errors while searching for orphaned bakes")
+        Logger.warn(s"${errors.length} errors exceeds the limit so orphaned bake deletion will not continue")
+      case _ =>
+        val bakes = Bakes.scanForAll()
+        val orphanedBakeIds = MarkOrphanedBakesForDeletion.findOrphanedBakeIds(recipes.map(_.id).toSet, bakes)
+        // TODO: minimal safety check to avoid accidentally deleting all the bakes if there are db read errors
```
jacobwinch: 😅

app/data/Recipes.scala
```diff
@@ -28,6 +30,19 @@ object Recipes {
     }
   }
 
+  def listWithErrors()(implicit dynamo: Dynamo): (List[DynamoReadError], List[Recipe]) = {
```
jacobwinch: What about _recipesWithErrors_?

app/data/Recipes.scala
```diff
@@ -28,6 +30,19 @@ object Recipes {
...
+    val dbResponse: Traversable[Either[DynamoReadError, DbModel]] = table.scan().exec()
+    val results = dbResponse.toList.separate
```
jacobwinch: I would use:

`val (errors, models) = dbResponse.toList.separate`

(Because I hate ._1 and ._2!)

app/controllers/HousekeepingController.scala
```diff
@@ -0,0 +1,33 @@
...
+import data._
+import housekeeping.MarkOrphanedBakesForDeletion
+import models.BakeId
+import play.api.Logger
+import play.api.mvc._
+
+class HousekeepingController(val authConfig: GoogleAuthConfig)(implicit dynamo: Dynamo) extends Controller with AuthActions {
+  private val (errors, recipes) = Recipes.recipesWithErrors
+  private val recipeIds = recipes.map(recipe => recipe.id).toSet
+  private val orphanedBakes = MarkOrphanedBakesForDeletion.findOrphanedBakeIds(recipeIds, Bakes.scanForAll())
```
sihil: Does this get initialised once when the application starts and never updated again?

app/housekeeping/MarkOrphanedBakesForDeletion.scala
sihil: What check needs to be done here?

katebee: If getting the recipes errors, or returns an empty list for some reason (because it kept getting DynamoDB read errors) then we are finding our orphaned bakes based on incomplete data... so housekeeping might fail to realise that a bake is actually in use

I think the FAULT_TOLERANCE partially addresses this... so I should have removed this TODO 😁 
```
case _ if errors.length > MarkOrphanedBakesForDeletion.FAULT_TOLERANCE =>
    Logger.info(s"Housekeeping found ${errors.length} database errors while searching for orphaned bakes")
    Logger.warn(s"${errors.length} errors exceeds the limit so orphaned bake deletion will not continue")
```

app/controllers/HousekeepingController.scala
katebee: Good point! Will fix

PR #215, opened 2018-08-09
Fix creation of kong dir

tomrf1:
I'm not sure why this stopped working, but adding this step fixes the issue in CODE


ajwl:
👍  confusing.. but glad it works.

PR #217, opened 2018-08-30
Improve "show bake" page

adamnfish:
The auto-scroll logic queues the animations rather than scrolling to the bottom immediately. This makes the page 100% unusable because after many 10s of updates have come through you're waiting for
literally minutes, unable to scroll the output while the jQuery animation queue completes. Without the queue this actually follows the output and you'll never have to wait more than `$animationDuration` (currently 500ms) before you can regain control.

Labelling the checkbox `tail -f` makes it unclear whether this refers to scrolling the div, or actually obtaining the log contents. It makes it feel like unticking this would stop the log loading at all. Auto scroll makes it clear we're just talking about the interface.

### DIFF
--- a/app/views/showBake.scala.html
+++ b/app/views/showBake.scala.html
@@ -29,7 +29,7 @@
   <div>
     <label>
       <input id="tail-log" type="checkbox" value="" checked>
-      tail -f
+      Auto scroll
     </label>
   </div>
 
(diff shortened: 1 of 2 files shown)


sihil:
@AWare this will conflict with #194 so you'll probably want to rebase.

adamnfish:
Has this become a "hit merge first" race?

PR #218, opened 2018-08-31
Role fixes to support Ubuntu 18.04

adamnfish:
Ubuntu 18.04 LTS (Bionic) introduces some changes that are not compatible with our existing roles.

This PR fixes the following problems:

## AWS Tools cannot be installed as-is

`easy_install` is no longer available on Ubuntu 18.04 but it's fine to use `pip` to install the package. Unfortunately the package depends on a library that does not work under Python 3 so we're forced to use `pip` rather than `pip3`. This means on Debian we need to ensure that pip is installed.

This change is backwards compatible with existing Linux versions, `pip` has been around for ages and installs packages just fine (which is why `easy_install` is no longer available).

## SSM comes pre-installed

Since the SSM Agent is already installed on the base image the manual `dpkg` install steps in our `ssm` role will fail. ~This change adds a condition to those commands so this operation is a no-op on Ubuntu 18.04. We could have left it as-is and told people "this role isn't required on 18.04 and your bakes will fail if you enable it" but that feels like Yzma's lair.~

Turns out the SSM role was actually broken / deprecated because all our base images include the ssm agent already. The change now removes the SSM role completely, avoiding Yzma's lair.

wrong-lever


adamnfish:
Turns out the ssm role was already deprecated. I've removed it from the repository.

adamnfish:
The role is now removed everywhere so I'm going to go ahead and merge this to fix the aws-tools installation on 18.04.

rtyley:
I'm having trouble `ssm`-ing into an Ubuntu Bionic image - I don't need to explicitly include the Amigo `ssm` role, right?

```
$ ssm ssh -x --profile ophan -i i-05e06c74a7d41270c
The specified instance(s) are not eligible targets (AWS said InvalidInstanceId)
bash: -c: option requires an argument
```

This is my recipe, not sure what I'm doing wrong: https://amigo.code.dev-gutools.co.uk/recipes/ophan-time-machine-test

rtyley:
Ah, Simon pointed out that I'd forgotten to assign the SSM role, d'oh, that fixed it and I can now ssm in :goat:

PR #221, opened 2018-10-17
new parameters for the 'node' role to allow downloading from nodejs.org

twrichards:
added additional parameters for the 'node' role to allow downloading from nodejs.org (rather than the current unofficial nodesource.com)

paired with @philmcmahon

### DIFF
--- a/roles/node/README.md
+++ b/roles/node/README.md
@@ -4,4 +4,14 @@ Installs Node.js.
 
 ## Variables
 
-* `node_version` - which version of Node.js to install. Default value is `4.x`. Valid values include `4.x` and `6.x`.
+#### Preferred Approach
+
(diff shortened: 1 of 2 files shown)


roles/node/README.md
```diff
@@ -4,4 +4,12 @@ Installs Node.js.
 
 ## Variables
 
-* `node_version` - which version of Node.js to install. Default value is `4.x`. Valid values include `4.x` and `6.x`.
\ No newline at end of file
+* `node_version` [DEPRECATED] - which version of Node.js to install (from nodesource.com - unofficial). Default value is `4.x`. Consult https://github.com/nodesource/distributions for valid values.
+
+#### OR Preferred Approach
```
jacobwinch: I would move this to the top if it's the preferred approach

mchv:
Downloading from `nodejs.org` rather than `nodesource.com` is a good security improvement 👍

PR #222, opened 2018-10-18
Puppeteer - Headless Chrome Node API

rtyley:
Ophan's Time Machine (https://dashboard.ophan.co.uk/time-machine) needs `Puppeteer` in order to be able to take full-page web screenshots - it used to use `phantomjs` but that is a dead project...

https://github.com/guardian/ophan/issues/2952

### DIFF
--- a/roles/puppeteer/tasks/main.yml
+++ b/roles/puppeteer/tasks/main.yml
@@ -1,0 +1,15 @@
+---
+- name: Install dependencies required by "puppeteer"
+  apt:
+    state: present
+    name: [
+      'npm', # we need this to *install* Puppeteer
+      'nodejs', # we need this to *run* Puppeteer
+      'chromium-browser' # easiest way to get all the X11 libraries required by Puppeteer's bundled copy of Chrome
+    ]
+
+- name: Install "puppeteer" node.js package so that it is available to any program on the system
+  npm:
+    name: puppeteer
+    version: '1.9.0'
+    path: / # Hack based on https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders - blame Alex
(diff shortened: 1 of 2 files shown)


roles/puppeteer/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Install Node - Ubunutu Bionic has a sufficiently recent version for puppeteer
+  apt:
+    name: nodejs
+    state: present
```
philmcmahon: could you maybe use the `node` role and set it as a dependency rather than installing it here?

roles/puppeteer/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+      'npm', # we need this to *install* Puppeteer
+      'nodejs', # we need this to *run* Puppeteer
+      'chromium-browser' # easiest way to get all the X11 libraries required by Puppeteer's bundled copy of Chrome
+    ]
+
+- name: Install "puppeteer" node.js package so that it is available to any program on the system
+  npm:
+    name: puppeteer
+    version: '1.9.0'
+    path: / # Hack based on https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders - blame Alex
```
rtyley: After fighting node's module resolution algorithm for about a day, I gave in when @AWare made this absolutely horrific suggestion that happens to work. I blame you fully Alex!

rtyley:
This role now seems to have `puppeteer` installed in a way that I should be able to run the updated Time Machine screenshotting code in https://github.com/guardian/ophan/pull/2959, so I'm... happy...

PR #224, opened 2018-10-23
Support adding keys from keyserver, extra docs

philmcmahon:
Packages such as http://pm2.keymetrics.io/docs/usage/install-as-deb/ require the key to be installed from the ubuntu keyserver. This change adds the option of specifying a key in this way, rather than via a url.

It also adds some documentation for testing amigo roles with vagrant, something I found saved a lot of time recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ it possible to SSH onto the instance that is being used to build the AMI.
 
 ## How to run locally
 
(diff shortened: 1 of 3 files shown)


philmcmahon:
That would be nice, it gets really messy once you're going back and forth between the vagrant friendly version and the amigo version

PR #225, opened 2018-10-23
Allow node role to install npm packages

philmcmahon:
This adds a step to add a symlink to the node `bin` directory (containing node, npm etc.) in /usr/local so that it can be referenced easily (without knowing the node version).

It also adds the functionality to allow the role to install npm packages globally. Arguably this belongs in a separate role, but due to the dependency between the [npm ansible step](https://docs.ansible.com/ansible/devel/modules/npm_module.html) and `npm` it is bit of a pain to separate this out. This doesn't feel like a terrible place to put this step - as it is optional, and I suspect will be used fairly often to install node and [pm2](http://pm2.keymetrics.io/).

Finally, I've removed the 'alternative' option of specifying the nodejs.org relative path on the basis that it adds quite a bit of complexity, and will be easy to add back in should it be required. Since nobody is using it at the moment we might as well keep things simpler.

Testing done: I spun up an EC2 box with an AMI generated with this role. I ran `node --version` and got the expected version.

### DIFF
--- a/roles/node/README.md
+++ b/roles/node/README.md
@@ -12,6 +12,8 @@ Using this approach the binary will be unpacked to `/usr/local/node` and a `node
 
 * `packages` - npm packages to install globally. e.g. `packages: [pm2, reqwest]`
 
+Using this approach, node binaries (node, npm, npx) will be accessible at `/usr/local/node/`. The `node` binary is linked to in `/usr/bin` so should be available via the `PATH` environment variable.
+
 #### DEPRECATED Approach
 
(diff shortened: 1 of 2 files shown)


twrichards:
`systemd` requires an absolute path for `ExecStart=` which without the symlink will need to be the unpack directory (which contains the node version, not ideal) - so please could we leave the symlink in?

see https://github.com/guardian/manage-frontend/pull/141/commits/da6f4e8196e24acaa171a3d2d943388c4dc4ccef

philmcmahon:
Nice spot, thanks! I've added it back in

twrichards:
@rtyley you might be interested in this

PR #237, opened 2018-12-24
Fix bakes using `node_version`

mbarton:
#225 missed a check that broke backwards compatibility with existing builds that simply specified `node_version`.


mbarton:
Ninja Christmas approval @jacobwinch, thanks!

I'm going to give it a test on CODE before merging

PR #239, opened 2019-01-17
Fix #238 by stopping the JVM caching DNS lookups forever.

### LINKED ISSUE #238: Java runtimes using SecurityManagers unpragmatically cache DNS hits *forever*
This is the default behaviour of Java runtimes (in both Java 8 & 11 apparently), defined in the `java.security` file - they cache DNS hits forever:

```
# The Java-level namelookup cache policy for successful lookups:
#
# any negative value: caching forever
# any positive value: the number of seconds to cache an address for
# zero: do not cache
#
# default value is forever (FOREVER). For security reasons, this
# caching is made forever when a security manager is set. When a security
# manager is not set, the default behavior in this implementation
# is to cache for 30 seconds.
#
# NOTE: setting this to anything other than the default value can have
#       serious security implications. Do not set it unless
#       you are sure you are not exposed to DNS spoofing attack.
#
#networkaddress.cache.ttl=-1
```

Although intended as a security measure against [DNS-spoofing](https://en.wikipedia.org/wiki/DNS_spoofing), it's not very realistic in the AWS-based-world in which we operate - services _do_ change their IP addresses from time to time, and so [AWS recommend](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-jvm-ttl.html) that a **`networkaddress.cache.ttl=60`** setting be applied.

On the Ophan team @amyhughes & I have seen this [result in us losing monitoring data](https://github.com/guardian/ophan/issues/3113) as the Elastic-hosted monitoring cluster changed it's IP without our Elasticsearch nodes noticing- they attempted to continue sending data to the old IP addresses.

Elastic themselves [recommend](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/networkaddress-cache-ttl.html) setting `networkaddress.cache.ttl` appropriately.

See also https://stackoverflow.com/q/1256556/438886

### The `java.security` configuration file

Locating and patching the `java.security` file is a little fiddly - this is where you find it on today's images:

```
/etc/java-8-openjdk/security/java.security
```

Could we bake the updated setting into the AMI role?

maxspencer:
Co-authored-by: Roberto Tyley <roberto.tyley@gmail.com>

### DIFF
--- a/roles/java8/tasks/main.yml
+++ b/roles/java8/tasks/main.yml
@@ -22,3 +22,9 @@
 - name: Install Java certificates
   command: /var/lib/dpkg/info/ca-certificates-java.postinst configure
   when: ansible_os_family == "Debian"
+
+## This modifies the JVM's DNS cache TTL, changing it from the default of INFINITY to 60
+## seconds. See this issue for full details: https://github.com/guardian/amigo/issues/238
+- name: Change JVM DNS cache TTL
+  command: sed -i'.backup' 's/#networkaddress.cache.ttl=.*/networkaddress.cache.ttl=60/g' $(dirname $(dirname $(readlink -f $(which javac))))/jre/lib/security/java.security
+


roles/java8/tasks/main.yml
```diff
@@ -22,3 +22,9 @@
 - name: Install Java certificates
   command: /var/lib/dpkg/info/ca-certificates-java.postinst configure
   when: ansible_os_family == "Debian"
+
+## This modifies the JVM's DNS cache TTL, changing it from the default of INFINITY to 60
+## seconds. See this issue for full details: https://github.com/guardian/amigo/issues/238
+- name: Change JVM DNS cache TTL
+  command: sed -i'.backup' 's/#networkaddress.cache.ttl=.*/networkaddress.cache.ttl=60/g' $(dirname $(dirname $(readlink -f $(which javac))))/jre/lib/security/java.security
```
alexduf: Wouldn't using $JAVA_HOME make the command easier to read? (Assuming it's defined by the ubuntu package)

maxspencer: We thought about that but it wasn't defined unfortunately. We considered a few other ways of finding the right path including `locate` and `dpkg` but settled on this in the end - it is a bit arcane though!

rtyley: We got that expression for getting the Java home folder from https://stackoverflow.com/q/1117398/438886 - you'll notice that they say `$JAVA_HOME` is now deprecated... :cry:

adamnfish: You can use any character as the delimiter for sed. I prefer not to use `/` when dealing with paths because it can be ambiguous whether it's a delimiter or a path part. This is a super minor point and these are filenames not paths so it doesn't even really apply, just an FYI.

alexduf: sounds good then!

rtyley:
This worked well with https://amigo.code.dev-gutools.co.uk/recipes/ubuntu-xenial-java8/bakes/35, not going to deploy to PROD:


> [2019-01-18 09:41:23] ubuntu-xenial-java8: TASK [java8 : Change JVM DNS cache TTL] ****************************************
> [2019-01-18 09:41:24] ubuntu-xenial-java8: changed: [127.0.0.1] => {"backup_file": "/etc/java-8-openjdk/security/java.security.32439.2019-01-18@09:41:24~", "changed": true, "msg": "1 replacements made"}

roles/java8/tasks/main.yml
davidfurey: If you split this into two steps, you can use `command` to get the path and then use this path as part of the `replace` step

PR #240, opened 2019-01-31
SBT 1.2.8 to teamcity-agent

gr211:


nicl:
have added @philmcmahon for review. I'd personally be really interested to see if this can improve build performance of Frontend.

philmcmahon:
I've tested this by running a build of dotcom (which uses the teamcity SBT plugin) on an agent running this updated version of the teamcity agent role. It worked! Let's try on PROD

PR #242, opened 2019-02-13
Airflow role

gr211:
Amigo Airflow role


roles/airflow/defaults/main.yml
```diff
@@ -0,0 +1,217 @@
...
+
+# MAIL
+airflow_email_backend: airflow.utils.email.send_email_smtp
+
+# SMTP
+airflow_smtp_host:
+airflow_smtp_starttls: True
+airflow_smtp_ssl: False
+airflow_smtp_port: 587
+airflow_smtp_mail_from: airflow@airflow.com
```
mchv: This does not look correct

roles/airflow/templates/airflow-add-connection.sh.j2
```diff
@@ -0,0 +1,36 @@
...
+CONNECTION_NAME=`echo $BASENAME`
+
+if [ -z $CONNECTION_NAME ]; then
+        echo "Could not figure connection name. It should be the remaining words after the first '_'"
+        exit 1
+fi
+
+echo "Creating connection $CONNECTION_NAME with type $CONNECTION_TYPE"
+
+JSON_ONE_LINE=`cat $FILE | sed 's/"/\\"/g' | perl -pe 's/\n//'`
```
mchv: not sure why you need `perl` to remove all end of lines

roles/pip/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: install pip
-  apt: name={{item}} update_cache=yes
-  with_items:
-    - python-pip
+  apt:
+    name: ['python-pip']
+    update_cache: yes
```
mchv: How those changes are related to `airflow`?

roles/efs-ubuntu/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
```
mbarton: Maybe add a README describing this role specifically?

roles/pip-packages/tasks/main.yml
```diff
@@ -1,4 +1,5 @@
 ---
 - name: Install a package with pip
-  pip: name={{ item }} state=present
-  with_items: "{{ packages }}"
\ No newline at end of file
+  pip:
```
mbarton: The changes look fine, just wondering what the reasoning was for them?

roles/packages/tasks/main.yml
```diff
@@ -34,8 +34,7 @@
   - ansible_os_family == "Debian"
 
 - name: Install a package with yum
-  yum: name={{ item }} state=present
-  with_items: "{{ packages|default([]) }}"
+  yum: name="{{ packages|default([]) }}" state=present
```
mbarton: What's the reason to change this?

roles/pip/tasks/main.yml
gr211: Old style is deprecated

roles/packages/tasks/main.yml
gr211: Old style is deprecated (and displays warnings)

roles/efs-ubuntu/defaults/main.yml
gr211: I do indeed intend to write one. There will be elements to be mentioned about `UserData` and such, and the minimum set values to be provided in order to get a functional airflow instances.

I would need to try it though to get the readme correct. Is there a way to test a new role in the context of aws ? Otherwise, I'll just write one readme and revisit it later.

roles/airflow/templates/airflow-add-connection.sh.j2
gr211: Mostly out of habit. I find perl more consistent in its regex engine for line breaks. And it is installed in the base ami.
I could replace it with `tr '\n' ''` though

roles/airflow/defaults/main.yml
gr211: Not sure I follow. These are just defaults

roles/airflow/templates/airflow-add-connection.sh.j2
mchv: @gr211 Fair enough. No need to change to use `tr`

roles/pip-packages/tasks/main.yml
gr211: There was there also some deprecation warnings. 
I also made `state` a variable to allow for other states beside `present`. The use case i had was that `setuptools` was already installed on the base image but to a version that caused some error with airflow's installation. The pb was that it was too old a version of that package.
Having `state == present` was not enough to get it to update.

roles/efs-ubuntu/defaults/main.yml
gr211: Done. Will revisit once tried with the real thing though.

PR #246, opened 2019-02-14
Airflow role - airflow service to be stopped during ansible build

gr211:


TBonnin:
Should the README be updated to mention that even though the airflow service is installed it is not started automatically

PR #304, opened 2019-05-20
Propagate prism errors and check for staleness of response

sihil:
AMIgo has been deleting AMIs that were in use due to a race condition at startup. If AMIgo runs a housekeeping job before it has managed to populate data from Prism it will think that no AMIs are in use as the instance and launch config lists will be empty. This is *not good*™️.

This PR:
 - introduces `Attempt` into the codebase and uses it to propagate error conditions from the Prism API (this wasn't strictly necessary but the codebase is on Scala 2.11 and no cats so no right biased either)
 - changes the initial state of the prism caches from `[]` to `Attempt.Left(NotInitialised)`
 - never sets the data cache back to `[]` when it can't parse the prism response
 - will not use stale data from prism
 - internally flags prism data as stale when we've not successfully updated in 15 minutes
 - modifies housekeeping and controller routes to deal with the resulting attempts that are propagated to the edge

### How to review
I'd strongly recommend starting at `app/prism/Prism.scala` and `app/services/PrismAgents.scala` as everything cascades from there. You might well want to load it into a real editor.

### DIFF
--- a/app/attempt/Failure.scala
+++ b/app/attempt/Failure.scala
@@ -1,0 +1,52 @@
+package attempt
+
+import java.io.{ PrintWriter, StringWriter }
+
+sealed trait Failure {
+  def msg: String
+  def cause: Option[Throwable] = None
+
+  def toThrowable: Throwable = cause.getOrElse(new RuntimeException(msg))
+}
+
+/**
+ * This type of failure has a throwable which could potentially be logged
+ */
+sealed trait FailureWithThrowable extends Failure {
+  def throwable: Throwable
+  override def cause = Some(throwable)
+
+  // provide a default mechanism for showing the exception to a user
+  override def msg = {
+    val stringWriter = new StringWriter()
+    val writer = new PrintWriter(stringWriter)
+    throwable.printStackTrace(writer)
+
+    stringWriter.toString
+  }
+}
+
+case object PrismNotInitialisedFailure extends Failure {
+  val msg = "Prism is not yet initialised"
+}
+
+case class PrismStaleFailure(msg: String) extends Failure
+
+case class JsonParseFailure(msg: String) extends Failure
+
+case class HousekeepingFailure(msg: String) extends Failure
+
+case class ConfigurationFailure(msg: String) extends Failure
+
+case class UnknownFailure(throwable: Throwable) extends FailureWithThrowable
+
+case class AwsSdkFailure(throwable: Throwable) extends FailureWithThrowable
+
+object Failure {
+  def collect[A](eithers: List[Either[Failure, A]])(recurse: A => List[Failure]): List[Failure] = {
+    eithers.flatMap {
+      case Left(failure) => List(failure)
+      case Right(success) => recurse(success)
+    }
+  }
+}
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -116,7 +116,7 @@ class AppComponents(context: Context)
   val prismAgents = new PrismAgents(prism, applicationLifecycle, actorSystem.scheduler, environment)
 
   // do this synchronously at startup so we can set permissions
-  val accountNumbers: Seq[String] = Await.result(prism.findAllAWSAccounts(), 30 seconds).map(_.accountNumber)
+  val accountNumbers: Seq[String] = Await.result(prism.findAllAWSAccounts().asFuture, 30 seconds).right.get.map(_.accountNumber)
 
   val sns: SNS = {
     val snsClient = AmazonSNSClientBuilder.standard
--- a/app/controllers/RecipeController.scala
+++ b/app/controllers/RecipeController.scala
@@ -1,5 +1,6 @@
 package controllers
 
+import attempt.Attempt
 import com.gu.googleauth.GoogleAuthConfig
 import data._
 import models._
@@ -12,6 +13,7 @@ import prism.RecipeUsage
 import schedule.BakeScheduler
 import services.PrismAgents
 
+import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.Try
 
 class RecipeController(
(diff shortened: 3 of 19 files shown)


app/attempt/Failure.scala
```diff
@@ -0,0 +1,52 @@
...
+
+case class JsonParseFailure(msg: String) extends Failure
+
+case class HousekeepingFailure(msg: String) extends Failure
+
+case class ConfigurationFailure(msg: String) extends Failure
+
+case class UnknownFailure(throwable: Throwable) extends FailureWithThrowable
+
+case class AwsSdkFailure(throwable: Throwable) extends FailureWithThrowable
```
adamnfish: This could probably extend something a little more specific? The AWS SDK has its own failure hierarchy IIRC. Not an important point :-)

app/components/AppComponents.scala
```diff
@@ -116,7 +116,7 @@ class AppComponents(context: Context)
   val prismAgents = new PrismAgents(prism, applicationLifecycle, actorSystem.scheduler, environment)
 
   // do this synchronously at startup so we can set permissions
-  val accountNumbers: Seq[String] = Await.result(prism.findAllAWSAccounts(), 30 seconds).map(_.accountNumber)
+  val accountNumbers: Seq[String] = Await.result(prism.findAllAWSAccounts().asFuture, 30 seconds).right.get.map(_.accountNumber)
```
adamnfish: This `.get` looks suspicious?

app/controllers/RecipeController.scala
```diff
@@ -12,38 +13,48 @@ import prism.RecipeUsage
...
+    val usages: Attempt[Map[Recipe, RecipeUsage]] = RecipeUsage.forAll(recipes, findBakes = recipeId => Bakes.list(recipeId))(prismAgents, ec)
+    usages.fold(failure => InternalServerError(s"Prism error: $failure"), { u =>
+      Ok(views.html.recipes(recipes, u))
+    })
   }
 
-  def showRecipe(id: RecipeId) = AuthAction { implicit request =>
-    Recipes.findById(id).fold[Result](NotFound) { recipe =>
+  def showRecipe(id: RecipeId) = AuthAction.async { implicit request =>
+    Recipes.findById(id).fold[Future[Result]](Future.successful(NotFound)) { recipe =>
```
adamnfish: This nested fold looks unnecessary to me. Attempt has Sync / Async contructors so you can line the sync and async elements into a single for comp and fold the result into a `Future[Result]]` (recall `Attempt.fold` returns `Future[B]`)

app/controllers/RecipeController.scala
```diff
@@ -115,47 +126,55 @@ class RecipeController(
     })
   }
 
-  def showUsages(id: RecipeId) = AuthAction { implicit request =>
-    Recipes.findById(id).fold[Result](NotFound) { recipe =>
+  def showUsages(id: RecipeId) = AuthAction.async { implicit request =>
+    Recipes.findById(id).fold[Future[Result]](Future.successful(NotFound)) { recipe =>
```
adamnfish: Another nested fold. I find it's always good to line the effects and types up so the controller is a single for-comp. It makes it much clearer what each Action does and strongly incentivises the separation of integration (controller) from implementation (functions applied in the for comp). This in turn dramatically improves testability of the application.

adamnfish: All that said, Attempt was designed before the current crop of Future replacements existed. I think it's worth looking at the much more robust alternatives offered by Cats, as well as various others.

app/packer/PackerRunner.scala
```diff
@@ -24,23 +26,25 @@ object PackerRunner extends Loggable {
...
-
-    executePacker(bake, playbookFile, packerConfigFile, eventBus, debug)
+    for {
+      awsAccounts <- prism.accounts
+      awsAccountNumbers = awsAccounts.map(_.accountNumber)
+      _ = log.info(s"AMI will be shared with the following AWS accounts: $awsAccountNumbers")
+      packerVars = PackerVariablesConfig(bake)
+      packerBuildConfig = PackerBuildConfigGenerator.generatePackerBuildConfig(bake, playbookFile, packerVars, awsAccountNumbers)
+      packerJson = Json.prettyPrint(Json.toJson(packerBuildConfig))
+      packerConfigFile = Files.createTempFile(s"amigo-packer-${bake.recipe.id.value}", ".json")
```
adamnfish: I feel like this is an operation that could fail? Should it return an `Attempt`?

app/packer/PackerRunner.scala
```diff
@@ -24,23 +26,25 @@ object PackerRunner extends Loggable {
...
+      _ = Files.write(packerConfigFile, packerJson.getBytes(StandardCharsets.UTF_8)) // TODO error handling
```
adamnfish: Here again, should this be an Attempt?

app/prism/Prism.scala
```diff
@@ -1,34 +1,46 @@
...
+  private def findAll[T](path: String)(implicit r: Reads[Seq[T]]): Attempt[Seq[T]] = {
     val url = s"$baseUrl$path"
-    ws.url(url).get().map { resp =>
-      extractData[Seq[T]](resp.json).fold(error => {
-        log.warn(s"Failed to parse Prism response for GET $url. Status code = ${resp.status}, Error = $error")
-        Seq.empty[T]
-      }, t => t)
+    Attempt.fromFuture(ws.url(url).get()) {
+      case NonFatal(t) => UnknownFailure(t)
+    }.flatMap { resp =>
```
adamnfish: Here again, this looks like nesting that may not need to exist (although I'm only skimming)? `traverse` is often the answer in cases like this.

app/prism/RecipeUsage.scala
```diff
@@ -14,46 +17,53 @@ object RecipeUsage {
...
+    } yield {
+      val bakeUsage = amisInUse.map { ami =>
+        val maybeDirectBake = bakedAmiLookupMap.get(ami)
+        val maybeCopy = copiedAmiLookupMap.get(ami)
+        val bake = (maybeDirectBake, maybeCopy) match {
+          case (Some(directBake), None) => directBake
+          case (None, Some(copy)) => bakedAmiLookupMap(copy.copiedFromAMI)
+          case _ => throw new IllegalArgumentException("AMI ID provided neither direct nor copied")
+        }
+        val amiInstances = instances.filter(_.imageId == ami)
```
adamnfish: Here we have some logic baked into the for-comp, which means it's hard to test these bits. I try and keep the logic away from the 'IO-edges' of the program like this so that fiddly bits like filtering can be tested. Up to you where the value / cost trade-off sits here.

If you're interested in testing the integration of large for-comps like these (which additionally makes it possible to test embedded logic), then I've been meaning to try [ZIO's environment abstraction](http://degoes.net/articles/zio-environment) to address this use-case. You'd be at the cutting-edge of Scala research if you wanted to give it a spin (although likely to be far along the cost / value trade-off scale for now).

test/test/AttemptValues.scala
```diff
@@ -0,0 +1,104 @@
...
+        case cause: Throwable =>
+          throw new TestFailedException((_: StackDepthException) => Some(s"Attempt resulted in a thrown exception in the underlying Future"), Some(cause), getStackDepthFun("AttemptValues.scala", "result"))
+      }
+    }
+  }
+}
+
+object StackDepthExceptionHelper {
+
+  def getStackDepth(stackTrace: Array[StackTraceElement], fileName: String, methodName: String, adjustment: Int = 0) = {
```
adamnfish: gosh

app/attempt/Failure.scala
sihil: Don't think this is used so will delete.

app/components/AppComponents.scala
sihil: It's in the AppComponents startup flow so if this fails the app will fail to start. I'm OK with that and didn't want to write complex retry logic.

philmcmahon:
Nice work! It seems like some of the complexity added is as a result of making the prism api calls asynchronous rather than doing await result? I guess nicer than having random timeouts though

app/controllers/RecipeController.scala
sihil: This Attempt has a `toAttempt` implicit on option - I'll refactor these.

app/controllers/RecipeController.scala
sihil: Typically I've written an action handler that puts this edge logic into a single place and makes the controllers nicer. Not bothered here as the priority was making AMIgo not dangerous. Will flatten the folds though.

app/packer/PackerRunner.scala
sihil: I've not attempted to attemptify the application outside of the highly dangerous prism operations. PRs welcome 😄

app/prism/RecipeUsage.scala
sihil: I'm aiming for the quickest way to make AMIgo not dangerous and not concerned that the logic here winds up baked into the for comp.

test/test/AttemptValues.scala
sihil: copy and pasted from a protected package in scalatest. think it's only needed because we are on an ancient version of scalatest in this project.

sihil:
@philmcmahon is right. I'm going to redo this to throw exceptions rather than half heartedly propagate `Attempt` throughout the codebase.

PR #314, opened 2019-06-06
Include the AWS SAM cli tool in the teamcity recipe

sihil:
This is required for running the `sam` CLI tool during builds on teamcity. Don't think it will impact anything else so should be safe.


sihil:
Silly idea. Closing.

PR #316, opened 2019-06-12
Add role for Tshark 🦈 ⭐️ 

katebee:
apt install tshark! 🦈 ⭐️ 

sharks


philmcmahon:
Ah I think you might be able to just use https://amigo.gutools.co.uk/roles#packages instead ?

PR #322, opened 2019-07-01
Enable Airflow paths to be set in the  launch configuration

pvighi:
Currently the paths to connections, logs, plugins and dags are set in role params in Amigo. Doing it this way forces us to hard code `PROD` parameters in the recipe.
By reading them from env variables we can set them in our launch configuration and set them to different paths depending on whether it is a `CODE` or `PROD` stack.

This PR should be merged after the [PR](https://github.com/guardian/amigo/pull/321) that fixes the Ubuntu Bionic base image

### DIFF
--- a/roles/airflow/README.md
+++ b/roles/airflow/README.md
@@ -87,18 +87,20 @@ Pass the following values to your AMIgo recipe:
 
 ```yaml
 airflow_remote_logging: True
-airflow_remote_base_log_folder: "s3://some-bucket/"
 airflow_remote_log_conn_id: dummy #set to a dummy value if not relying on actual airflow connections but rather on ec2 permissions/role. Cannot be empty or null though.
 
 ```
-
+And set the following env variable in your `UserData` as previously explained
+```bash
+airflow_remote_base_log_folder=some-bucket-name/some-path
+```
 ## DAGs remote sources configuration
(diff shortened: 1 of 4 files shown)


app/models/BaseImage.scala
```diff
@@ -29,10 +29,10 @@ case object Ubuntu extends LinuxDist {
     PackerProvisionerConfig.executeRemoteCommands(Seq(
       // Wait for cloud-init to finish first: https://github.com/mitchellh/packer/issues/2639
       "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done",
-      "apt-get --yes install software-properties-common",
+      "DEBIAN_FRONTEND=noninteractive  apt-get --yes install software-properties-common",
       "apt-add-repository --yes ppa:ansible/ansible",
       "apt-get --yes update",
-      "apt-get --yes install ansible"
+      "DEBIAN_FRONTEND=noninteractive apt-get --yes install ansible"
```
pvighi: This are just the changes in the [bugfix PR](https://github.com/guardian/amigo/pull/321) I just created this branch from that one to be able to run this in `CODE` without it failing. We should merge that PR first

PR #333, opened 2019-09-02
Airflow: add shelljob to airflow pip installs

laurenemms:
Add shelljob to Airflow pip installs so that it is available for the new backfill plugin.


pvighi:
did you test it in `CODE` ?

laurenemms:
> did you test it in `CODE` ?

just tested :)

laurenemms:
closing this because we're not going to use the plugin that requires this dependency anymore.

PR #334, opened 2019-09-05
AWS config cleanup

fredex42:
adding cleanup script for phantom ".aws" directories left behind owned by root.

These have been present regularly on TC build agents and are breaking builds that rely on pulling a Docker image from ECR @philmcmahon @AWare


fredex42:
tested locally using the Vagrant method detailed in Amigo

AWare:
Maybe setuid would be safer?

fredex42:
setuid discussed with @AWare and we decided to leave it as-is

PR #336, opened 2019-09-09
Bump the version of airflow to the latest stable version 1.10.5.

JonNorman:
We are also removing the explicit flask dependency when installing airflow as we don't believe that this is necessary anymore.

There are a number of other dependencies that are included in the bake, which I'd question - e.g. `numpy` - but I'm not planning on doing a full audit of airflow's requirements today.

### DIFF
--- a/roles/airflow/defaults/main.yml
+++ b/roles/airflow/defaults/main.yml
@@ -6,7 +6,7 @@ nfs_mount_id:
 nfs_mount_point: /mnt/nfs
 
 ## General
-airflow_version: 1.10.3
+airflow_version: 1.10.5
 airflow_extra_packages: ['crypto,postgres,google_auth,s3']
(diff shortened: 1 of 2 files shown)


TBonnin:
This changes the version for all recipes using this role. I think we are the only one at the moment but it is better practice to minimize unforseeable impact. I think it is better to bump the version number in our recipe param.

JonNorman:
Sounds sensible but why not both update our recipe and update the default while I'm at it?

The point of having Amigo images is so that we can all use sensible base line images that are updated and current. What is the advantage of leaving an old version as a default (given I've guaranteed all other teams - i.e. no one - is happy with the change)?

TBonnin:
As long as we can guarantee that nobody else is using the role or this change won't impact them. go for it

JonNorman:
It's just us for now!

image
image

PR #339, opened 2019-10-01
fix bugs in cloudwatch-hardware-monitoring-cronjob role

Mullefa:
Issues:
- `when` indentation meant that condition wasn't used and therefore facts were always set (therefore, all command line options always provided)
- logic to join paths meant that first path in array would not be prefixed with `--disk-path` (which is why providing `--disk-path` explicitly works [here](https://amigo.gutools.co.uk/recipes/centralised-elk-lk))

When this is merged I will update the roles which depend on this recipe accordingly.

### DIFF
--- a/roles/cloudwatch-hardware-monitoring-cronjob/README.md
+++ b/roles/cloudwatch-hardware-monitoring-cronjob/README.md
@@ -1,0 +1,11 @@
+# CloudWatch hardware monitoring cronjob
+
+Utilises [mon-put-instance-data.pl](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/mon-scripts.html) script to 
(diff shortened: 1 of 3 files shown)


Mullefa:
FYI these are the roles dependent on this recipe; I'll update them once this PR has been deployed. There _may_ be a loss of metrics temporarily.

- [ ] datatech-airflow
- [ ] ubuntu-wily-discussion
- [ ] identity-base
- [ ] elk
- [ ] centralised-elk-lk
- [ ] ophan-elasticsearch
- [ ] xenial-java8-deploy-infrastructure
- [ ] centralised-elk-es
- [ ] ubuntu-xenial-ophan-data-lake
- [ ] ubuntu-xenial-discussion

PR #340, opened 2019-10-18
Add fastlane role for app distribution

jacobwinch:
This PR adds a role which installs [`fastlane`](https://docs.fastlane.tools/). 

A few teams (iOS, Android, Editions) are now using `fastlane` for app distribution, so I'd like to include this role on the TeamCity build agents if possible.

### DIFF
--- a/roles/fastlane/meta/main.yml
+++ b/roles/fastlane/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+  - install-ruby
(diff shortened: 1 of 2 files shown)


philmcmahon:
Looks good to me. I wonder whether it makes sense to make this as a generic 'install ruby gem' role though? Might save future people some time

jacobwinch:
Thanks @philmcmahon - closing this in favour of https://github.com/guardian/amigo/pull/344.

PR #345, opened 2019-11-06
Add Role for Wazuh Manager

katebee:
This should create an AMI with Wazuh Manager, Wazuh API and Filebeat installed...

This is the AMI that we will use for the Wazuh Cluster:

image

### DIFF
--- a/roles/wazuh-manager/tasks/main.yml
+++ b/roles/wazuh-manager/tasks/main.yml
@@ -38,8 +38,3 @@
   apt:
     name: filebeat
     update_cache: yes
-
-- name: Download base config file and alert template
-  shell: |
-      curl -so /etc/filebeat/filebeat.yml https://raw.githubusercontent.com/wazuh/wazuh/v3.9.2/extensions/filebeat/7.x/filebeat.yml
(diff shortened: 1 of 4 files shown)


roles/wazuh-manager/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    update_cache: yes
+
+- name: Install Wazuh API
+  apt:
+    name: wazuh-api
+    update_cache: yes
+
+- name: Disable the Wazuh updates
+  shell: |
+    echo "wazuh-manager hold" | sudo dpkg --set-selections
```
akash1810: does this mean we could miss out on security updates?

katebee: Good question! This is recommended by the Wazuh installation guide:

Screenshot 2019-11-08 at 12 13 31

> It is recommended that the Wazuh repository be disabled in order to prevent accidental upgrades

The main reason seems to be to avoid compatibility issues between the different systems.
InfoSec will need to decide a patching and upgrade schedule.

PR #351, opened 2020-02-14
Add role for new cloud watch agent

Mullefa:
Some of the functionality this brings is already provided by other Amigo roles, however, this is the canonical approach suggested by AWS for EC2 instance metric collection; allows for more configuration; and enables the collection of SDK metrics (c.f. the work I'm about to do - will report back when I've learnt more).

Tested, locally and on CODE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,7 @@ target
 /RUNNING_PID
 amigo.iml
 
+Brewfile.lock.json
+
 roles/*.log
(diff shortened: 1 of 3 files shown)


Mullefa:
Potentially yes. Though tbh there may not be much motivation to migrate if all an application wants is basic disk and memory metrics of the EC2 instance running it. I'll report back with any additional learnings.

PR #354, opened 2020-02-24
Use all disks for disk space monitoring

AWare:
Shell script courtesy of @aug24

### DIFF
--- a/roles/disk-available-cronjob/tasks/main.yml
+++ b/roles/disk-available-cronjob/tasks/main.yml
@@ -3,5 +3,5 @@
   cron:
     name: "disk space avaible"
     # Which minutes or how often (*/5)
-    minute: "{{ minute }}"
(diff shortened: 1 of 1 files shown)


roles/disk-available-cronjob/tasks/main.yml
```diff
@@ -1,4 +1,3 @@
 ---
 - name: Add disk available cronjob
-  cron: name="disk space avaible" minute="*/5" job="/usr/local/aws-scripts-mon/mon-put-instance-data.pl --disk-space-avail --mem-avail --disk-path /dev/xvda1 --auto-scaling --from-cron"
```
sihil: You've dropped the `minute="*/5"` here. My guess is that the default is to drop to 1 minute? I wonder that could be explicit or configurable.

AWare: Yeah, this should be a config option.

PR #356, opened 2020-02-27
Role: RDS Root CA Certificates

paddyo:
For anyone interested in using full on SSL when connecting to an RDS service. This role gets the certificate, adds it to a new truststore & imports the default Java truststore certificates into that one.

### DIFF
--- a/roles/aws-rds-root-cert-ca/defaults/main.yml
+++ b/roles/aws-rds-root-cert-ca/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+truststore_password: changeit
+castore_password: changeit
--- a/roles/aws-rds-root-cert-ca/defaults/main.yml
+++ b/roles/aws-rds-root-cert-ca/defaults/main.yml
@@ -1,3 +1,2 @@
 ---
-truststore_password: changeit
 castore_password: changeit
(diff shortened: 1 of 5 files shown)


roles/aws-rds-root-cert-ca/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+truststore_password: changeit
```
AWare: Do we want to set defaults at all for this, it seems more dangerous than letting it fail if they're not set?

sihil:
Have we thought about bundling with the the java roles? It makes sense to me that we should have java that just works. What are the pros and cons here? Have to maintain multiple versions for each java we support? This might be the case anyway as the truststore path might be different?

paddyo:
Yeah apparently the truststore paths are pretty arbitrary apparently. Also most people won't be using these RDS certs because the VPC will do the job for them.

sihil:
As in we don't consider necessary to ensure the link is secure within a VPC?

paddyo:
Yes, I believe VPC is considered secure for a lot of people (not sure if that's correct or not). We are definitely putting an extra layer on identity though because of personal information.

PR #357, opened 2020-02-28
Mt add service to install airflow python

tylr-mk:
Add a new service to Airflow to download and install the Python directory published by the new Riff Raff generated instance.

### DIFF
--- a/roles/airflow/templates/usr/local/bin/airflow-install-ophan-python.sh.j2
+++ b/roles/airflow/templates/usr/local/bin/airflow-install-ophan-python.sh.j2
@@ -1,0 +1,18 @@
+#!/bin/sh
+GUARDIAN_PYTHON_ZIP=guardian.zip
+PYTHON_FOLDER=python/
+S3_PATH="s3://${airflow_s3_dags_folder}/$PYTHON_FOLDER/$GUARDIAN_PYTHON_ZIP"
+LOCAL_PYTHON_FOLDER=
+
+echo "Downloading extra python libraries from $S3_PATH"
+rm -f {{ airflow_home }}/python
+mkdir {{ airflow_home }}/python
+aws s3 cp $S3_PATH {{ airflow_home }}/
+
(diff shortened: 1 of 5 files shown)


roles/airflow/templates/usr/local/bin/airflow-install-ophan-python.sh.j2
```diff
@@ -0,0 +1,38 @@
...
+# can then be a static bash script.
+echo "Downloading extra python libraries from $S3_PATH"
+# Brute force a removal of the folder whether or not it exists already
+rm -f {{ airflow_home }}/$PYTHON_FOLDER
+aws s3 cp $S3_PATH {{ airflow_home }}/
+
+if [ -e {{ airflow_home }}/$GUARDIAN_PYTHON_ZIP ]; then
+	# Inflate / extract the python installable, note that files will now
+	# be in airflow_home/python/guardian/ etc.
+    unzip -o {{ airflow_home }}/$GUARDIAN_PYTHON_ZIP -d {{ airflow_home }}
```
pvighi: would this break if `{{ airflow_home }}/guardian.zip` doesn't exist ?
 if so we should make sure we merge the [airflow PR](https://github.com/guardian/ophan-data-lake/pull/4947) quickly after this one or tweak one of the two PRs to be backwards compatible and merge that one first

tylr-mk: Nope... there are no system changes really... this is kind of a safe handover for the image as it were. Where it will do this step if it can, but there would be no structural impact if it cant.

The `-e` conditional in line::24 means that the unzip step will only be run if the zip file exists. And will just spew the WARNING to stdout (as is the pattern in other files) and continue with  the rest of the run.

tylr-mk: This however doesn't safely handle a failed unzip but then we have bigger problems!

PR #367, opened 2020-04-24
Fix quoting on string in disk available

AWare:
This was causing `grid-imaging` and `grid-imaging-eoan` to fail.
@aug24 @philmcmahon


AWare:
This builds on code.

PR #369, opened 2020-05-05
Use ubuntu default for ansible as ppa not updated.

AWare:
It doesn't look like ansible's ppa is going to support focal soon, and there's an official package for focal. https://packages.ubuntu.com/focal/ansible
https://github.com/ansible/ansible/issues/69203

This also updates the colour scheme to the iTerm2 default.


app/models/BaseImage.scala
```diff
@@ -30,7 +30,10 @@ case object Ubuntu extends LinuxDist {
       // Wait for cloud-init to finish first: https://github.com/mitchellh/packer/issues/2639
       "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done",
       "DEBIAN_FRONTEND=noninteractive  apt-get --yes install software-properties-common",
-      "apt-add-repository --yes ppa:ansible/ansible",
+      "version=$(. /etc/os-release; echo $VERSION_ID | cut -d'.' -f1)",
+      "if (($version < 20)); then apt-add-repository --yes ppa:ansible/ansible; fi",
```
mbarton, line 34: Should this be `version > 20`?

AWare: We only want to add the PPA before ubuntu focal

mbarton: Ah my mistake, I thought it was the other way round

AWare:
Afaict, `aws-tools` depends on python 2 and 3. As python 2 is no longer included in Ubuntu, I'm not entirely sure what to do. 

I'm not sure what the cloudformation helpers do, or what we need them for. If they're not commonly used, I might split them out into their own packages and leave aws-tools to just depend on them. (With something to deliberately stop it on ubuntu 20.04 maybe)

sihil:
Version 2 of the AWS CLI is entirely self contained so I reckon we switch aws-tools to use that. Either as a swap out for everyone or as an aws-tools-2 that grabs that.

sihil:
The AWS CFN helpers include the cfn-init tool which is the tool that parses and applies the `AWS::CloudFormation::Init` section. i.e. https://github.com/guardian/deploy-tools-platform/blob/master/cloudformation/elk.template.yaml#L370-L399

sihil:
I'd also check whether the ubuntu images now include the aws cli and aws cfn tools in the base image. The images became far more AWS friendly out the box like two years back now so could well be that we don't need this stuff at all - if we do there are likely to be OS packages for them which we could switch to.

AWare:
@adamnfish also suggests that the python 3 incompatibility may have since been resolved. I'll have an explore and see what I find.

AWare:
Using a workaround to use python 2 (hooray)
But ansible's documentation on os dependant tasks makes less than no sense to me https://ansible-tips-and-tricks.readthedocs.io/en/latest/os-dependent-tasks/variables/

roles/aws-tools/tasks/install-cfn-tools-Ubuntu-20.yml
```diff
@@ -0,0 +1,45 @@
...
+  - name: Download AWS CFN tools
+    get_url: url=https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz dest=/tmp/aws-cfn-bootstrap-latest.tar.gz 
+  
+  - name: Create directory for AWS CFN tools
+    file: path=/tmp/aws-cfn-bootstrap-latest state=directory
+  
+  - name: Extract AWS CFN tools
+    command: tar xvfz /tmp/aws-cfn-bootstrap-latest.tar.gz --strip-components=1 -C /tmp/aws-cfn-bootstrap-latest
+  
+    # this package has dependencies that are not compatible with Python 3 so we use `pip`
```
adamnfish, line 17: Just observing that it doesn't look like this actually uses pip anymore, so this comment no longer matches (this explicitly runs the setup.py script instead).

roles/aws-tools/tasks/install-cfn-tools-Ubuntu-20.yml
```diff
@@ -0,0 +1,45 @@
...
+    until: result.rc == 0
+    retries: 5
+  
+  - name: Find all the utilities
+    find:
+      paths: /usr/local/bin/
+      patterns: cfn-*
+    register: utils
+
+  - name: Replace the python 3 shebang with the 2
```
adamnfish, line 32: This might be more clearly worded as something like "switch utils to explicitly use python 2, now that python 3 is the default".

roles/aws-tools/tasks/install-cfn-tools-Ubuntu-20.yml
```diff
@@ -0,0 +1,45 @@
...
+    register: utils
+
+  - name: Replace the python 3 shebang with the 2 
+    replace: 
+      path: "{{ item }}"
+      regexp: "\/usr\/bin\/env python"
+      replace: "/usr/bin/env python2"
+    with_items: "{{ utils.files | map(attribute='path') | list }}"
+
+  # - name: "Rub temples and complain about python 3 (Replace shebang on cfn-tools with python2)"
```
adamnfish, line 39: If this isn't needed would it be better to remove it? It's in the history if we want to refer to it in the future.

roles/aws-tools/tasks/main.yml
```diff
@@ -1,27 +1,13 @@
...
-  until: result.rc == 0
-  retries: 5
-
-- name: Cleanup AWS CFN tools
-  file: path=/tmp/aws-cfn-bootstrap-latest state=absent
+- name: Install pip and the latest AWS CLI
+  include: "{{ item }}"
+  with_first_found:
+    - "install-pip-aws-cli-{{ ansible_distribution }}-{{ansible_distribution_major_version }}.yml"
+    - "install-pip-aws-cli-{{ ansible_os_family }}.yml"
```
adamnfish, line 6: Do we need to plan ahead for 11 months time when this breaks on version 21?

roles/aws-tools/tasks/install-cfn-tools-Ubuntu-20.yml
AWare: Yeah, I'll delete.

AWare:
The theme stuff is in a separate commit, just pulling that out into a new PR. Will tidy up the comments.

AWare:
https://github.com/guardian/amigo/pull/370

roles/aws-tools/tasks/main.yml
AWare: Yeah, this isn't good.

AWare: I'm not sure what would be better. I think we're going to need to revisit this several times as python 2 is removed and the amazon tools are updated.

sihil: I think this is fine for now. When we have more clarity on what AWS are doing with CFN tools it'll be easier to figure out how to make this generic enough for all versions.

sihil: Also - what does this do on Ubuntu < 20? Does `roles/aws-tools/tasks/install-pip-aws-cli-debian.yml` need to be renamed?

sihil: Is there really a difference between Ubuntu 20 and Ubuntu <20 for the CLI file? Ubuntu seems to install `python-pip` additionally but then not use it - wonder whether this split is really required.

AWare: pip2 is used to install the cloudformation tools on ubuntu <20

AWare:
I've changed things to target ubuntu 20+ instead of just 20. 
I did test this for Xenial, but will run again on this commit.

AWare:
This is building on >20 and <20 🎈

roles/aws-tools/tasks/main.yml
```diff
@@ -1,27 +1,17 @@
 ---
-- name: Install latest AWS CLI
-  include: install-pip-aws-cli-redhat6.yml
-  when: ansible_os_family == "RedHat"
+- name: See what Ubuntu version we're running
+  set_fact: 
+    post_focal: "{{ 'post-focal' if ( ansible_distribution_major_version|int >= 20 ) else 'older' }}"
```
sihil, line 4: niggle is that this isn't really `post-focal` as it includes focal. but whatever...

AWare: this was really hard to find a "good" name for.

PR #371, opened 2020-05-14
safely mount ebs volumes

AWare:
With Amazon nitro, EBS volumes are made available at paths like `/dev/nvme0n1`. 
There's no strict dependency between what the device is and it's path. 
They have turned out to mostly be deterministic, however the new Ubuntu version changes that. 

This uses the script from @mbarton  and @joelochlann to find the correct device path from the supposed device path that is set in the EBS. I've taken the liberty of ripping out half the script and adding the UUID to the fstab per amazons documentation. 

*TODO*
- [ ] build this into a nitro instance and test it (including reboot)
- [ ] build this into a not nitro instance and test it (including reboot)

### DIFF
--- a/roles/ebs-volume/README.md
+++ b/roles/ebs-volume/README.md
@@ -1,6 +1,8 @@
 # Attach EBS Volume
 
 A script to mount ebs volumes, because it's not safe to assume a device path.
+On aws "nitro" instances, NVM is used, and the mapping between nvm device path
+and EBS path is non-deterministic. 
 
 
 Usage: attach-ebs-volume -d device-letter -m mountpoint [-u user]
--- a/roles/ebs-volume/README.md
+++ b/roles/ebs-volume/README.md
@@ -1,0 +1,25 @@
+# Attach EBS Volume
+
+A script to mount ebs volumes, because it's not safe to assume a device path.
+
+
+Usage: attach-ebs-volume -d device-letter -m mountpoint [-u user]
+
+  This script creates and attaches an encrypted EBS volume.  
+
+    -d dev-letter The device letter. This should be a single character (usually
+                  h or later) that is used to identify the device. Note that the
+                  device name specified by Amazon and understood by Ubuntu are
+                  different.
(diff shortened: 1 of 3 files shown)


roles/ebs-volume/README.md
```diff
@@ -0,0 +1,25 @@
+# Attach EBS Volume
+
+A script to mount ebs volumes, because it's not safe to assume a device path.
```
mbarton, line 3: Nitpick: it's worth expanding a bit here in the README that it's not safe to assume a path because the mapping between the device name as seen in EBS and the NVM device name on the machine is non-deterministic

roles/ebs-volume/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+  
+  - name: copy script to mount drive
+    copy: 
+      src: attach-ebs-volume.sh
+      dest: /opt/attach-ebs-volume/attach-ebs-volume.sh 
+      owner: root 
+      group: root
+      mode: u+x
+
+  - name: symlink into /usr/local/sbin
```
mbarton, line 16: Why not copy straight into `sbin`? (probably a newbie question)

AWare: Don't know, was cribbing off of `mount-drive`

roles/ebs-volume/README.md
AWare: Good point 👍

roles/ebs-volume/files/attach-ebs-volume.sh
```diff
@@ -0,0 +1,154 @@
...
+  return 0
+}
+
+function find_uuid() {
+  blkid -s UUID -o value $UBUNTU_DEVICE
+}
+
+function find_fstab_id() {
+  local UUID=$(find_uuid)
+  if [[ -z "${UUID// /}" ]]; then
```
mbarton, line 88: I've not seen `// /` before (bash n00b). What does it do?

AWare: space stripping

PR #374, opened 2020-07-09
Docker credentials and Amazon Linux 2 base image

nicl:
*Note: have successfully built an Amazon Linux 2 AMI with the `docker-ecr` role on Amigo CODE and then tested that in the Frontend AWS account.*

## What does this change?

We use multiple languages and the surrounding tooling (cloudformation, AMIs, riffraff bundling, etc.) is language-specific. This means a lot of repetition. Docker provides a standard interface here.

Amazon Linux 2 has been used as the base image for a few reasons, but primarily because it has good out-of-the-box support for Docker (with the `amazon-linux-extras` package).

Re-use of deployment code across languages/projects.

## Questions

`UserData` in cloudformation runs as `root`, but I've added the credentials helper config to the `ec2-user` home, which means it will only be found when launching a docker command with the ec2-user user.

What do people think on this? Should I assume people will only run docker as root?

### DIFF
--- a/roles/docker-ecr/tasks/main.yml
+++ b/roles/docker-ecr/tasks/main.yml
@@ -1,0 +1,58 @@
+---
+- name: Enable Docker
+  command: amazon-linux-extras install docker
+
+- name: Install Docker
+  yum:
+    name: docker
+    state: present
+
+- name: Add ec2-user to docker group to avoid requiring 'sudo'
+  user:
+    name: ec2-user
+    groups: docker
+    append: yes
+
+- name: Install Amazon ECR Docker Credential Helper
+  yum:
+    name: amazon-ecr-credential-helper
+    state: present
+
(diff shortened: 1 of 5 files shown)


roles/docker-ecr/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: Create Docker config directory (if doesn't exist)
+  file:
+    path: /home/ec2-user/.docker
+    state: directory
+    mode: ugo+rw
+    owner: ec2-user
+    group: ec2-user
+
+- name: Copy Docker config to register creds store
```
nicl, line 29: I should just repeat this and the previous step for `/root` to ensure it works for both the `root` and `ec2-user` users.

amyhughes:
Fantastic!

PR #375, opened 2020-08-04
Add nest-secrets recipe

nicl:
## What does this change?

Adds a [nest-secrets](https://github.com/guardian/nest-secrets) role.


nicl:
@mbarton yeah a get_url role which adds something to the path might be nice!

PR #378, opened 2020-08-19
Add a link to view the AMI in AMIable.

akash1810:
Easiest to review w/out whitespace https://github.com/guardian/amigo/pull/378/files?w=1.

## What does this change?

Deep-link into AMIable. This view in AMIable provides information about the AMI, including:
- creation date
- ARN
- any launch configurations that use it
- any instances that use it
- any available updates

This is useful to double check if we have setup a stack's auto-update correctly: if we see the most recent AMI not used, but an AMI from last month is, then we've likely misconfigured the app in RiffRaff.

NOTE: Although there is a CODE stack for AMIgo, there is only a PROD stack for AMIable.

This also makes the structure of the `table` a bit simpler with a dedicated column for the copy button and the new AMIable link, whereas before the copy button was part of the ami id column.

## How to test

- Deploy to CODE
- Open a recipe's recent bake page
- Click the new `View in AMIable` link to open AMIable

Tighter integration in our tooling, making discovery easier.

Leaking of the AMIable domain, however its been mentioned that the domain is in the AMIable repo already.

## Images

**Before**
image

**After**
image

### DIFF
--- a/app/views/showRecipe.scala.html
+++ b/app/views/showRecipe.scala.html
@@ -87,6 +87,7 @@
               <button class="btn btn-primary btn-xs absolute-right copy-button" title="Copy to clipboard" data-clipboard-text="@bake.amiId">
                 <img src="@routes.Assets.versioned("images/clippy.svg")" width="13" alt="Copy to clipboard">
               </button>
+              <a href="https://amiable.gutools.co.uk/ami?imageId=@bake.amiId" target="_blank" rel="noopener noreferrer">View in AMIable</a>
             }
             </td>
           </tr>
@@ -109,6 +110,7 @@
                 <button class="btn btn-primary btn-xs absolute-right copy-button" title="Copy to clipboard" data-clipboard-text="@copy.imageId">
                   <img src="@routes.Assets.versioned("images/clippy.svg")" width="13" alt="Copy to clipboard">
                 </button>
+                <a href="https://amiable.gutools.co.uk/ami?imageId=@bake.amiId" target="_blank" rel="noopener noreferrer">View in AMIable</a>
               </td>
             </tr>
           }
--- a/app/views/showRecipe.scala.html
+++ b/app/views/showRecipe.scala.html
@@ -71,45 +71,78 @@
           }
           <th>Status</th>
           <th>AMI</th>
+          @* Column for copy button *@
+          <th></th>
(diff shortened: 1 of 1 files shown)


app/views/showRecipe.scala.html
```diff
@@ -71,45 +71,80 @@ <h4>Inherited from base image @recipe.baseImage.id</h4>
...
+                </button>
+              </td>
+              <td>
+                <a href="https://amiable.gutools.co.uk/ami?imageId=@bake.amiId" target="_blank" rel="noopener noreferrer">View in AMIable</a>
+              </td>
+              </td>
+            } else {
+              <td></td>
+              <td></td>
+              <td></td>
```
adamnfish, line 118: Does this `else` clause need an extra column to account for the new Amiable col?

adamnfish (approved):
Hard to read the diff with all the formatting changes, but broadly looks sensible to me, excepting one possible markup error.

Also:

Does it still work ok on smaller screens (down to a reasonable size, I guess)?

If that's a problem, I often try and stick to the mantra that "text != UI", which might help keep it working at the smaller sizes. Maybe there's a button icon that would work instead of using words to describe the UI - ideally with something like "view in Amiable" as tooltip / label still for discoverability and accessibility!

This is not a very important point, and maybe we don't care about Amigo working on smaller screens in any case :-)

app/views/showRecipe.scala.html
akash1810: I think three is correct. This code block will run when there is no `amiId`.

If there is an `amiId` we render three columns:
1. ami id
1. copy button
1. amiable link

So three empty `td`s are correct here, I think.

akash1810:
> Does it still work ok on smaller screens (down to a reasonable size, I guess)?

Good shout. This is it at 598px (the smallest Chrome will resize to). It looks useable to me. (thanks bootstrap!)

image

> Maybe there's a button icon that would work instead of using words to describe the UI - ideally with something like "view in Amiable" as tooltip / label still for discoverability and accessibility!

Small screens are typically mobiles. How one views tooltips on mobiles, though possible, isn't obvious imo. If we were to go down the icon route, I'd push for an image that represents AMIable, for example the AMIable logo as it makes the destination obvious.

PR #382, opened 2020-10-12
Make docker pull via nexus.gutools

jorgeazevedo:
## What does this change?

This configures the docker daemon to pull via nexus rather than straight from docker hub. This is to avoid failures due docker hub rate limiting.

## How to test

A new spot instance that runs a teamcity build that triggers a docker pull should have an effect on nexus. Docker images should start to be cached at https://nexus.gutools.co.uk/#browse/browse:docker

N/A

If this is misconfigured, docker pulls will fail. So every teamcity build that relies on it will grind to a halt.

### DIFF
--- a/roles/teamcity-agent/tasks/main.yml
+++ b/roles/teamcity-agent/tasks/main.yml
@@ -30,6 +30,8 @@
     dest: /usr/local/bin/docker-compose
     mode: 0555
     checksum: sha256:f679a24b93f291c3bffaff340467494f388c0c251649d640e661d509db9d57e9
+- name: set docker to use nexus pull through proxy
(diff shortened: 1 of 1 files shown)


jorgeazevedo:
Currently Blocked by https://github.com/guardian/deploy-tools-platform/pull/236

philmcmahon (approved):
Nice! 

relevant whale gif

jorgeazevedo:
@philmcmahon I've shipped the changes in nexus, so this is good to go. I'm not going to be around for the rest of the week so feel free to ship this, alternatively we can pick up when I come back Monday next week.

PR #383, opened 2020-10-15
add a link to GitHub search

akash1810:
## What does this change?

AMIgo already provides details of load balancers and instances running an AMI built by a recipe, which is great. But we don't know which repository the apps run in.

This PR adds a link to a GitHub search, searching for references of the recipe id within `riff-raff.yaml` files in the GitHub organisation makes it easier to know which repos use a particular recipe.

This allows us to find which apps might be affected when editing a recipe.

The URL will go to a search like [this](https://github.com/search?q=org:guardian+xenial-java8-deploy-infrastructure+filename:riff-raff.yaml&type=code).

**Example use case**
I'm looking to deprecate the `s3-ssh-keys` role as ssm has replaced it.

This role was also causing the `xenial-java8-deploy-infrastructure` recipe to [fail to bake](https://amigo.gutools.co.uk/recipes/xenial-java8-deploy-infrastructure/bakes/359).

We can already see where a role is used within AWS. Knowing which repo a recipe is used allows us to confirm that the cloudformation template has the required resources defined to enable ssh over ssm and thus allow the safe removal of the `s3-ssh-keys` role from the recipe.

## How to test

- Deploy to CODE
- Open recipe
- Click link

Success is determined by the resulting GitHub search being useful.

It's easier to find a recipe's use in VCS.

n/a

## Images

image

### DIFF
--- a/app/views/showRecipe.scala.html
+++ b/app/views/showRecipe.scala.html
@@ -37,6 +37,11 @@
       } else {
         <p><b>This recipe is not used</b></p>
       }
+      <p>
+        <a href="@(s"https://github.com/search?q=${helper.urlEncode(s"""org:guardian+"${recipe.id}"+filename:riff-raff.yaml""")}&type=code")">
+          Search GitHub for usage
+        </a>
+      </p>
     </div>
   </div>
 


app/models/Recipe.scala
```diff
@@ -13,7 +14,12 @@ case class Recipe(
   modifiedBy: String,
   modifiedAt: DateTime,
   bakeSchedule: Option[BakeSchedule],
-  encryptFor: List[AccountNumber])
+  encryptFor: List[AccountNumber]) {
+
+  // Arguably this belongs in the template,
```
adamnfish, line 19: I think the fact that this comment exists is a strong sign that this isn't the right place for this code :-)

What isn't working about the template, I agree that this sounds like a concern of the presentation rather than the model. If we really must have it in the Scala codebase itself rather than the template, then better to keep it separate from the data models themselves by not adding it as a method. I'd prefer to see this as a function on an object in that case. As a point of interest, written that way that function would only need to take the `RecipeID`, for ease of testing.

adamnfish (commented):
What an excellent idea!

I agree with your comment though :-)

shtukas (approved):

app/models/Recipe.scala
akash1810: I think I've got it working 🎉 .

I wasn't wrapping the expression in `@` previously, which seems to have been the issue. I find Twirl a little confusing, especially after writing React for most client side ui!

app/views/showRecipe.scala.html
```diff
@@ -37,6 +37,11 @@ <h1>@recipe.id.value</h1>
       } else {
         <p><b>This recipe is not used</b></p>
       }
+      <p>
+        <a href="@(s"https://github.com/search?q=${helper.urlEncode(s"""org:guardian+"${recipe.id}"+filename:riff-raff.yaml""")}&type=code")">
```
akash1810, line 41: `"""` to avoid escaping the `"` surrounding the recipe id in the URL.

PR #385, opened 2020-10-26
Add new oauth2-proxy role for version 6.x

katebee:
## What does this change?

Add a new role that will make it possible to install OAuth2 Proxy version 6.x

Rather than altering the existing role, this PR creates a new one.

Steps were adapted from the existing role and have been somewhat tested as part of the [InfoSec Wazuh stack](https://github.com/guardian/infosec-platform/blob/508e46c70ef4a732645e404e8f19a8decab4b9f9/cloudformation/wazuh/wazuh-cluster.yaml#L655).

## What is the value of this and how can we measure success?

The existing role is called `google-auth-proxy` and relies on a [GitHub repository](https://github.com/bitly/oauth2_proxy) that is now archived. 

The new role will download oauth2-proxy from the [official fork](https://github.com/oauth2-proxy/oauth2-proxy) which is being maintained and developed:

- The last release of `bitly/oauth2_proxy` was v2.2 in April 2017. 
- The last release of `oauth2-proxy/oauth2-proxy` was v6.1.1 in August 2020

Since OAuth2-Proxy can be used for Google, Github and other providers, it seemed reasonable to use a name that did not specifically reference only one provider. This also allows us to avoid updating the current role (which is in use by at least one application) which reduces the risk of releasing and testing this, and of breaking authentication for systems that rely on this role. 

😅 🤞 😅 🤞 

## How to test

Good question! We can either merge this and see if we can build an AMI with the new role that works as expected.
Otherwise, we can try out the Vagrant and Virtualbox route for local testing.

By not switching over the existing Oauth2 Proxy role we ensure existing recipes will continue to bake function AMIs for the applications that use them. Once we have verified this role works we can switch over and remove the old role.

##### 🦇 🧙‍♀️ 🎃 HAPPY HALLOWEEN 🎃 🧙‍♀️ 🦇

trick-pacman

### DIFF
--- a/roles/oauth2-proxy/defaults/main.yml
+++ b/roles/oauth2-proxy/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+oauth_version: 6.1.1
+systemd_file: oauth2-proxy.service
--- a/roles/oauth2-proxy/files/oauth2-proxy.service
+++ b/roles/oauth2-proxy/files/oauth2-proxy.service
@@ -1,0 +1,15 @@
+
+[Unit]
+Description=Oauth2 Proxy
+
+[Service]
+Type=simple
+User=oauth2-proxy
+Group=oauth2-proxy
+EnvironmentFile=-/etc/default/oauth2-proxy
+ExecStart=/opt/oauth2-proxy/oauth2-proxy --config @@OAUTH2_CONFIG_PATH
+Restart=on-failure
+RestartSec=3
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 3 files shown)


philmcmahon (approved):

philmcmahon:
Oauth2 into the future! halloween

akash1810:
Is it worth planning how to remove the `google-auth-proxy` role from existing recipes?

katebee:
@akash1810 there is only [one recipe](https://amigo.gutools.co.uk/recipes/centralised-elk-lk) using `google-auth-proxy`.

I have just tested the new role with [wazuh.gutools.co.uk](wazuh.gutools.co.uk) and it works okay 👍 
Regarding migration, I would be keen to merge this so that `centralised-elk-lk` can be switched to version 6+ asap.

One way to ensure this happens and track progress would be for me to open a draft PR removing `google-auth-proxy` with a TODO list that we can tackle before the PR can be merged.

akash1810 (approved):

PR #386, opened 2020-10-28
Delete the google-auth-proxy role

katebee:
## What does this change?

https://github.com/guardian/amigo/pull/385 added a new role that will make it possible to install OAuth2 Proxy version 6.x

The `google-auth-proxy` role relies on a [GitHub repository](https://github.com/bitly/oauth2_proxy) that is now archived. 
Deleting this role will prevent the creation of recipes that rely on an unmaintained and outdated version of Oauth2 Proxy.

## What needs to be done before this is merged?

- [x] Switch [the recipe](https://amigo.gutools.co.uk/recipes/centralised-elk-lk) that uses this role to use the oauth2-proxy role
- [x] Build an AMI for `centralised-elk-lk` using the updated recipe
- [x] Check if we need to update the oauth2-proxy [configuration file](https://github.com/guardian/deploy-tools-platform/blob/master/cloudformation/elk.template.yaml#L583) for Central ELK
- [x] Deploy Central ELK and ensure that OAuth2 based authentication still works still works

1. Improved security of our applications
2. Delighted InfoSec team 🧙‍♀️✨

This should only be merged once we have switched Central ELK to using the oauth2-proxy role.

Central ELK currently using command-line arguments to configure OAuth2 Proxy. These may still work when we update Central ELK to using version 6.x of OAuth2 Proxy. However, it is currently 4 major versions behind so we might have some breaking changes to deal with as part of the upgrade. 🤷


katebee:
There are currently no recipes using [the google-auth-proxy role](https://amigo.gutools.co.uk/roles#google-auth-proxy).
We should check again before merging but this should be ready to review 👀 🙏

philmcmahon (approved):
💥

PR #390, opened 2020-11-18
download the shared ssh key onto the teamcity agent

akash1810:
## What does this change?

This is to allow installation of packages from private github repositories via git+ssh.

See https://github.com/guardian/deploy-tools-platform/pull/280.

## How to test

- Deploy to CODE
- Run a bake of TeamCity agent
- Observe it succeed

TeamCity agents are able to interact with private git repositories.

n/a

## Images

n/a

### DIFF
--- a/roles/teamcity-agent/tasks/main.yml
+++ b/roles/teamcity-agent/tasks/main.yml
@@ -74,6 +74,18 @@
   become_user: teamcity
   args:
     executable: /bin/bash
+- name: add ssh key
+  shell: |
+    set -e
+    mkdir -p /opt/teamcity/ssh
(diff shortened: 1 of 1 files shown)


philmcmahon (approved):

sihil (changes requested):
I think we need to have further discussion about putting secrets onto baked AMIs. Such AMIs are not encrypted and by default are shared with all Guardian AWS accounts which is wider sharing than I feel totally comfortable with and certainly a lot wider than the key in S3 as it stands today.

Better approach might be grabbing the data from SSM at agent boot time if feasible?

Would like to discuss further before merging.

akash1810:
Closing in favour of the more idiomatic #391 .

PR #392, opened 2020-11-23
Add Wazuh Agent role

jorgeazevedo:
## What does this change?

This adds a new role called wazuh-agent. It installs the `wazuh-agent` package and patches its systemd unit so it calls `authenticate-with-wazuh-manager.sh` _before_ the actual agent runs.

`authenticate-with-wazuh-manager.sh` patches wazuh-agent's config file so it connects properly, and registers the box using `agent-auth`

The role is parameterised and requires a number of variables to run, namely

* manager_ip
* agent_version
* secret_arn

For reference: bash version https://github.com/guardian/amigo/blob/f8c1745bd4ee96172dd26a11a3665324e0a2cd69/roles/wazuh-agent/tasks/main.yml

### Notes on systemd

systemd is the way modern ubuntu manages services. Each service has an associated text file that describes its implementation according to systemd's interface. These include properties like what commands do you run to start and stop it, which other services it depends on etc. More importantly for our purposes, the systemd interface also includes a notion of a **prestart command**, which we can use as a hook to call the authentication script before the agent proper starts. Here's a minimal example

```
      [Unit]
      After=network.service

      [Service]
      ExecStartPre=authenticate.sh
      ExecStart=start-agent.sh

      [Install]
      WantedBy=default.target
```

systemd has a really neat interface based on `systemctl` for actions and `journalctl` for logs . These are useful for debugging

```
# Print current status
systemctl status wazuh-agent

# Print full log
journalctl -u wazuh-agent
```

## How to test
Deploy to CODE and launch an instance with the ami. That's enough to test that we can add it to the base ubuntu image though?

TBD

## Outstanding:
* Set -e on /usr/local/bin/authenticate-with-wazuh-manager.sh . Requires handling the aws failure on lack of permissions to describe tags
* Remove jq, it's already installed

### DIFF
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -19,20 +19,15 @@
 
     sed -i "s/^deb/#deb/" /etc/apt/sources.list.d/wazuh.list
     apt-get update
-    echo installed_agent
+    
     # ----------------- Store the password and register the agent
 
     SECRET_ARN="arn:aws:secretsmanager:{{ secret_region }}:{{ account_id }}:secret:{{ secret_id }}"
-    echo $SECRET_ARN
-
     REGION=`curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone | sed 's/.$//'`
-    echo fetched_region
 
     aws secretsmanager get-secret-value --secret-id $SECRET_ARN \
        --query SecretString --output text --version-stage AWSCURRENT --region $REGION \
        | jq -r .value > /var/ossec/etc/authd.pass
-
-    echo fetched_secret
 
     cat << 'PARENT_EOF' > /usr/local/bin/connect-to-wazuh-manager.sh
     #!/bin/bash -x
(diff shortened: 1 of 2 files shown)


philmcmahon (approved):

philmcmahon:
Param wise, maybe we could switch it to try pass the whole arn of the secret as a param so that there's just 1 instead of 3 params?

philmcmahon:
One thing we need to remember is that we'll need to make  sure instances have the 'describe tags' permmission so that we get the app/stack/stage tags

jorgeazevedo:
> Param wise, maybe we could switch it to try pass the whole arn of the secret as a param so that there's just 1 instead of 3 params?

Yeah, good point! I'll try to have another go at setting a value with `:`

jorgeazevedo:
@philmcmahon found it! The right way to escape it is to use single quotes

roles/wazuh-agent/tasks/main.yml
```diff
@@ -0,0 +1,89 @@
...
+
+- name: Inject authentication script
+  copy:
+    dest: /usr/local/bin/authenticate-with-wazuh-manager.sh
+    mode: '0744'
+    content: |
+      #!/bin/bash -x
+      if [ -f "/var/ossec/etc/authd.pass" ]; then
+        MANAGER_IP="{{ manager_ip }}"
+        REGION=`curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone | sed 's/.$//'`
```
philmcmahon, line 53: it might be worth giving `169.254.169.254` a variable name so that it's clear that it's not just a magic IP address

philmcmahon (approved):
Nice one

roles/wazuh-agent/tasks/main.yml
jorgeazevedo: Good point! Pushed a commit to address that

PR #393, opened 2020-12-01
Allow amigo instance to communicate with wazuh

jorgeazevedo:
## What does this change?

This adds the amigo instance to a new security group that allows it to connect to wazuh on ports 1414 and 1415.

## How to test
Deploy to code!


jorgeazevedo:
Oh. This actually isn't needed as outbound traffic is already unrestricted 🤦‍♂️

PR #395, opened 2020-12-08
Upload list of installed packages to S3 at the end of a bake

philmcmahon:
## What does this change?
This adds a new step to *all* bakes on amigo which uploads a list of installed packages to a bucket specified in configuration. The uploaded file is tagged with two tags, one with the recipe ID, the other with the bake ID. Using these two parameters we can associate the package list with AMIs in amiable. This can help us answer questions such as 'what software is installed on discussion API instances' and 'are we using X library anywhere?' (harder, would require full download of the bucket contents).

## How to test
Run a bake with amigo (this is on CODE at the moment) - you should see a final step where the package list is uploaded to S3, and see the resulting file in S3. 

With a tape measure. The longer the package list the better. Anything over 130cm counts as victory.

This change introduces a new thing to keep track of and clean up. Strictly speaking there should be a cleanup job just like we have for AMIs to delete these package lists once the associated AMI has been deleted. I'm lazy though, and these text files are pretty small.


jorgeazevedo:
> With a tape measure. The longer the package list the better. Anything over 130cm counts as victory.

😂

conf/application.conf
```diff
@@ -25,3 +25,5 @@ ansible {
     # s3prefix: "an/optional/prefix/"
   }
 }
+
+amigo.data.bucket = "bucket-to-save-packagelist-to"
```
jorgeazevedo, line 29: So how does this work? Does this file have placeholder values, and the prod app gets its conf from S3?

philmcmahon: Amigo uses a (now legacy) library called configuration magic which magically replaced Play config with config in dynamo https://github.com/guardian/configuration-magic - so this value is set there. I've included it here so that it's obvious what to override locally

sihil:
Questions:
 - The upload happens directly from the instance if I understand correctly, if so then we need to add `s3:PutObject` permission in the CFN policy for the packer instances
 - Can/should we create the S3 bucket in the CFN?
 - Can we have an example in the doc about how to fetch the files for all in-use bakes?
   - Do we need an API endpoint to get list of live recipe/bake-id tuples?

philmcmahon:
Yep we'll need to add that permission. I'm currently using a pre-existing bucket called amigo-data which appears to be orphaned currently. We could potentially adopt it, or create a separate one - it's currently also used for storing scripts that are required by some ansible roles

conf/application.conf
jorgeazevedo: ah, thanks! You mentioned in another comment which bucket you've used so far. I was curious to see what the data looks like, I've found an example now!

jorgeazevedo:
Re-using amigo-data seems ok to me, there's 2 subdirectories on there already, this would be a third. Unless we'd need to apply different retention/permission rules?

I guess that leads to a question I had, I've understood what questions this data can answer for us, but I haven't understood the use case itself. Is the idea to have this data available to enable manual lookups via something like amiable?

philmcmahon:
I'm beginning to think that initially this data will just be in amigo - we could maybe display it underneath the build log or instead of the build log for succesful bakes (e.g. on this page https://amigo.gutools.co.uk/recipes/bionic-java8/bakes/98). In terms of doing any serious querying of it, that's possibly out of scope for now. Maybe an immediate use case would be someone seeing 'oh this AMI is 100 days old, I wonder if it has X insecure package installed?' - even having the package list available on an AMI by AMI basis could be useful, as if the package exists on one app, it's likely to exist on every instance using that AMI, and possibly every instance with the same base image.

philmcmahon:
Though if I'm honest, the main motiviation is something that came up in a couple of our meetings with Infosec so it's a bit of a tick box exercise - I don't know exactly how it will be used. my currenty plan is to add some kind of UI to view the package list to Amigo and then stop and see what Kate/Stili think about it, or if further features are required

sihil:
Get the permission in place so it actually works and then we can card something up to create the MVP of "give me all of the files". Be good to have the AMI ID in the file but guess that's not possible as it's not known when this is uploaded.

philmcmahon:
I've updated this so that the package list is exposed on an API endpoint (/recipe/bake/packages) and in the UI:

Screenshot 2020-12-15 at 12 33 09

philmcmahon:
@sihil unfortunately I couldn't think of an easy way to get the AMI ID into the file without retrospectively adding it as the file is made during the bake which is before the AMI id gets generated

philmcmahon:
Depends on https://github.com/guardian/deploy-tools-platform/pull/314

ansible/playbook.yaml
```diff
@@ -4,5 +4,5 @@
   connection: local
   become: true
 
-  tasks:
-    - name: Your task(s) go here...
\ No newline at end of file
+  roles:
+    - role-to-test
```
philmcmahon, line 8: this is unrelated but I think a useful change for people trying to test roles using Vagrant

philmcmahon:
I think this is in a good state to merge. I've addressed some issues raised by @jorgeazevedo  before christmas (I hope I got them all as I was storing theem in my head) - mainly restructuring the upload command and improving the error reporting when amigo fails to fetch a package list.

This PR gives the packer user the required permissions to perform the upload https://github.com/guardian/deploy-tools-platform/pull/320

Thanks to @sihil's comments I've decided to cloudform the amigo-data bucket and have a separate bucket for code/prod. All package list data will go to the new `amigo-data-stage` buckets. I'll migrate the existing data from `amigo-data` to `amigo-data-prod` and delete `amigo-data` following this PR.

TODO in a different PR: provide an easy way to get all the package lists for currently in use AMIs

jacobwinch (approved):
The code looks good to me 👍 

I think it makes sense to merge this as a first pass and then add further improvements as we learn how InfoSec / other teams will use this functionality.

PR #396, opened 2020-12-08
WIP: Add BakeInstance parameter to BaseImage to allow use of arm64 images.

philmcmahon:
## What does this change?
This introduces a new propety to the 'base image' type called BakeInstance. This will determine the AWS instance type that is used to bake AMIs for that base image. This is currently hard coded to `t3.small` but following the [release of t4g instances](https://aws.amazon.com/ec2/instance-types/t4/) we might want to be running the `arm64` architecture at the guardian. 

## How to test
Deploy to amigo CODE, create base image with arm64 and an arm64 base AMI, try to perform a bake. Currently on AMIgo you'll get an error: `Error launching source instance: InvalidParameterValue: The requested instance type's architecture (x86_64) does not match the architecture in the manifest for ami-0346ee471bb2c892a (arm64)` - this should resolve that

being able to bake arm64 images at the Guardian

Introducing an entirely new archiecturecould have unexpected side effects so we should roll this out carefully if we decide to go down the arm64 route.

Screenshot 2020-12-08 at 19 27 25

### DIFF
--- a/app/controllers/BaseImageController.scala
+++ b/app/controllers/BaseImageController.scala
@@ -29,7 +29,8 @@ class BaseImageController(
       val form = Forms.editBaseImage.fill((
         image.description,
         image.amiId,
-        image.linuxDist.getOrElse(Ubuntu)
+        image.linuxDist.getOrElse(Ubuntu),
+        image.bakeInstance.getOrElse(BakeInstances.x86)
       ))
       Ok(views.html.editBaseImage(image, form, Roles.listIds))
     }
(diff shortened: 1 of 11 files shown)


sihil:
I would expect that we should be able to look up the architecture of the provided base image AMI and figure out which instance type we should use without making users have to figure it out themselves.

sihil:
Be good to chat about this tomorrow. 

I imagine that people will effectively want to clone existing recipes and get ARM bakes. That then led me to wonder about the whole flow - how does riff-raff know which AMI to pick. Should it figure that out from the launch configuration or do we just name the image `_arm` all the way through? Will we want to continuously bake x86 and arm of a given recipe for any length of time or will a recipe only need to target one?

philmcmahon:
closing in favour of https://github.com/guardian/amigo/pull/397

PR #399, opened 2020-12-10
Cloning recipes and base images

sihil:
## What does this change?

It is frequently the case that a new recipe is needed that is similar to a prior recipe. The same is true of base images. This PR adds the ability to close both of these and also adds the ability to delete unwanted and unused base images (you can already do this for recipes).

## How to test

When deployed to CODE you should be able to clone both recipes and base images and delete both when unused. You should be unable to delete a base image that is used by recipes.

## Images

image


philmcmahon (approved):
:shipit: LGTM

katebee:
Hooray! 👍 🥳  This will be a huge help, thanks!

Noticed a small issue if you click `clone` without entering a name for the clone:

Screenshot 2021-01-13 at 11 00 06
I know what I did wrong so maybe not worth addressing? 🤷

sihil:
Ummm, not super helpful error! I've raised #524 which might fix this (I've not tested it yet).

PR #400, opened 2020-12-21
Add pip3 role to support Ubuntu 20.04

mbarton:
The package in Focal is `python3-pip` even though there's no Python 2 in that release. Before this PR, the `pip-packages` role didn't work on that release as it hard-codes a dependency on `python-pip`.

This PR adds two new roles, `pip3` and `pip3-packages` that can be used on both Focal and Bionic to install Python 3 packages. As Python 2.7 is EOL we would expect users to migrate to using these roles and we can eventually remove the old ones. The naming however makes it clear which version of Python you are expecting to use, regardless of if the distribution packages 2.7 or not (thanks @AWare for the suggestion).

In fact, most of our AMIs already have Python 3 pip. We install it as as a dependency of [aws-tools](https://github.com/guardian/amigo/blob/a6323b56801b5880d683ff293bb74fb600595a6b/roles/aws-tools/tasks/install-pip-aws-cli-Debian.yml#L3). So in the actual bake the `pip3` role install is a no-op but we have kept the package around for consistency. It feels sensible for `pip-packages` to depend on `pip`, even if in reality our all images also include the `aws-tools` role already.

### DIFF
--- a/roles/pip/tasks/main.yml
+++ b/roles/pip/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: install pip
   apt:
-    name: ['python-pip']
+    name: ["{{ 'python3-pip' if (ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int >= 20) else 'python-pip' }}"]
     update_cache: yes
     state: latest
--- a/roles/pip/tasks/main.yml
+++ b/roles/pip/tasks/main.yml
@@ -1,8 +1,8 @@
 ---
 - name: Check Ubuntu version
-    fail:
-      msg: Ubuntu 20.04 (Focal) and above no longer package Python 2. Use the `pip3` role instead.
-    when: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int >= 20
+  fail:
+    msg: Ubuntu 20.04 (Focal) and above no longer package Python 2. Use the `pip3` role instead.
+  when: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int >= 20
 
 - name: install pip
   apt:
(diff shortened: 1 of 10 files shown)


roles/pip/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: install pip
   apt:
-    name: ['python-pip']
+    name: ["{{ 'python3-pip' if (ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int >= 20) else 'python-pip' }}"]
```
mbarton, line 4: I *think* (but haven't tested) that the package in RHEL is still called `python-pip`. No RHEL bake is using this role atm anyway

shtukas (approved):

roles/pip/tasks/main.yml
akash1810, line 4: Could we add an inline comment to help future travellers understand this?

AWare:
I can only find a few usages of pip in:
https://amigo.gutools.co.uk/roles#pip
https://amigo.gutools.co.uk/roles#pip-packages
and none of the usages are particularly clear about why they depend on it. 

I'm slightly concerned that jumping python 2-3 will cause difficult to spot failures if the eventual use of pip is to install a python 2 module. 

I would be half tempted to just mark these as not supporting focal and then add a python 3 set which only support focal+. Given the not particularly wide use of python, I don't think we'd want to go further than that.

roles/pip/tasks/main.yml
mbarton: We've decided to add separate roles for Python 3 which hopefully provides the clarity a comment would have had here

mbarton:
@AWare I've updated this PR to add new roles for `pip3` and `pip3-packages`

AWare (approved):
Aside from ansible arcana to prevent using it on the wrong ubuntu version, this is perfect.

mbarton:
Arcana added 👍

PR #402, opened 2021-01-07
Use correct amigo-data bucket in roles. Delete mongo role

philmcmahon:
amigo-data is now amigo-data-prod - see https://github.com/guardian/amigo/pull/395 for details. There's also an amigo-data-code bucket. This creates some confusion as if the language-tool-ngrams role (the only one which accesses amigo-data) runs on Amigo CODE it will still query the amigo-data-prod bucket, which doesn't make much sense, but since this bucket contains other data that is useful to split between CODE/PROD, this awkward situation seems reasonable. A cleaner solution would be for thte langauge-tool-ngrams role to detect which stage it's running on, but that would add unnecessary complexity to the role.

We no longer use Mongo DB at the guardian, so I've deleted those roles.

### DIFF
--- a/roles/language-tool-ngrams/defaults/main.yml
+++ b/roles/language-tool-ngrams/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
-source_tarball: s3://amigo-data/data/language-tool/ngrams.tar.gz
+source_tarball: s3://amigo-data-prod/data/language-tool/ngrams.tar.gz
 target_dir: /opt/ngram-data


akash1810:
Could we split this PR in two to keep to single responsibilities? It makes for an easier revision history for future developers.

philmcmahon:
Thanks for resurrecting this Akash!

roles/language-tool-ngrams/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-source_tarball: s3://amigo-data/data/language-tool/ngrams.tar.gz
+source_tarball: s3://amigo-data-prod/data/language-tool/ngrams.tar.gz
```
akash1810, line 2: I'm a little confused with this change, and have some questions:
  1. How are recipes using this role currently working?
  2. Why are we not stage aware?
  3. Does CODE have access to this PROD bucket? Does this follow best practice?

akash1810:
Given this was raised ~2 years ago, and there have been no reported issues, I'm tempted to close this PR as not needed.

@philmcmahon @guardian/developer-experience any objections?

akash1810:
> Given this was raised ~2 years ago, and there have been no reported issues, I'm tempted to close this PR as not needed.
> 
> @philmcmahon @guardian/developer-experience any objections?

Closing.

PR #403, opened 2021-01-12
Add endpoint to list bake usages

philmcmahon:
## What does this change?
This adds a /bakeUsages endpoint to list all the bakes in AMIgo that are currently in use. One use for this would be to fetch package lists for all bakes currently in use.

As part of this changed I've added some tests for the messty string manipulation bits of the package list work
## How to test
I've tested this on CODE but it's not easy to test locally or on CODE as no locally or CODE baked images are 'in use' so you might have  to take my word for it

The endpoint works, and gets used at some point. 

## Images

Screenshot 2021-01-12 at 12 09 25

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -39,7 +39,7 @@ GET     /recipes/:recipeId/bakes/:buildNumber           controllers.BakeControll
 GET     /recipes/:recipeId/bakes/:buildNumber/packages  controllers.BakeController.bakePackages(recipeId: RecipeId, buildNumber: Int)
 GET     /recipes/:recipeId/bakes/:buildNumber/events    controllers.BakeController.bakeEvents(recipeId: models.RecipeId, buildNumber: Int)
 
-GET     /bakeUsages                 controllers.BakeController.allBakeUsages
+GET     /bake-usages                 controllers.BakeController.allBakeUsages
 
 # Healthcheck
 GET     /healthcheck                controllers.RootController.healthcheck
--- a/conf/routes
+++ b/conf/routes
@@ -39,6 +39,8 @@ GET     /recipes/:recipeId/bakes/:buildNumber           controllers.BakeControll
 GET     /recipes/:recipeId/bakes/:buildNumber/packages  controllers.BakeController.bakePackages(recipeId: RecipeId, buildNumber: Int)
 GET     /recipes/:recipeId/bakes/:buildNumber/events    controllers.BakeController.bakeEvents(recipeId: models.RecipeId, buildNumber: Int)
 
+GET     /bakeUsages                 controllers.BakeController.allBakeUsages
+
 # Healthcheck
 GET     /healthcheck                controllers.RootController.healthcheck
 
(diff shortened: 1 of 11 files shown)


jacobwinch:
> This adds a /bakeUsages endpoint to list all the bakes in AMIgo that are currently in use.

I think we'd normally build AMI-auditing functionality into AMIable, rather than AMIgo. Is there a particular reason that we've decided to build this into AMIgo in this case?

conf/routes
```diff
@@ -39,6 +39,8 @@ GET     /recipes/:recipeId/bakes/:buildNumber           controllers.BakeControll
 GET     /recipes/:recipeId/bakes/:buildNumber/packages  controllers.BakeController.bakePackages(recipeId: RecipeId, buildNumber: Int)
 GET     /recipes/:recipeId/bakes/:buildNumber/events    controllers.BakeController.bakeEvents(recipeId: models.RecipeId, buildNumber: Int)
 
+GET     /bakeUsages                 controllers.BakeController.allBakeUsages
```
jacobwinch, line 42: Minor: I think this should be `bake-usages` (to fit with the convention used elsewhere in this file e.g.`base-images`)

test/data/PackageListTest.scala
```diff
@@ -0,0 +1,23 @@
+package data
+
+import models.{BakeId, RecipeId}
+import org.scalatest.{FlatSpec, Matchers}
+
+class PackageListTest extends FlatSpec with Matchers {
```
jacobwinch, line 6: 💯

test/models/BakeIdTest.scala
```diff
@@ -0,0 +1,16 @@
...
+class BakeIdTest extends FlatSpec with Matchers {
+
+  "toFilename" should "produce expected filename" in {
+    val bakeId = BakeId(RecipeId("puking-pastilles"), 1)
+    BakeId.toFilename(bakeId) should be ("puking-pastilles--1.txt")
+  }
+
+  "toMetadata" should "produce valid metadata" in {
+    val bakeId = BakeId(RecipeId("nosebleed-nougat"), 1)
+    BakeId.toMetadata(bakeId) should be ("Recipe=nosebleed-nougat,BuildNumber=1")
```
jacobwinch, line 13: Tremendous test data Strings @philmcmahon  ⚡

jacobwinch (approved):

philmcmahon:
> > This adds a /bakeUsages endpoint to list all the bakes in AMIgo that are currently in use.
> 
> I think we'd normally build AMI-auditing functionality into AMIable, rather than AMIgo. Is there a particular reason that we've decided to build this into AMIgo in this case?

There's no strong reason. My thought was that AMIable deals with all our AMIs whereas the intention here is just to list AMIs that AMIgo built, so it's reasonable to put the endpoint in AMIgo. 

The *actual* reason was that it was a fair bit easier doing it here than in AMIable!

PR #524, opened 2021-01-13
Add some min lengths for IDs

sihil:
## What does this change?
There is currently no check on whether an ID for a new (or cloned) base image or recipe has characters in it. The quickest solution is to add a minlength to the form.
 
## How to test

Try to create or clone a recipe with no name.

Better errors than a unknown exception explosion.


philmcmahon (approved):

katebee (approved):
👍  cheers

PR #526, opened 2021-01-20
Adds SSM capability to Packer

Nirvikalpa108:
## What is the problem?
Security HQ was showing an error when trying to collect the security groups of the Deploy Tools account. After debugging, we learnt that AWS Trusted Advisor was unable to return a response for an account with over 200 security groups. The Deploy Tools account had over 850! Most of these were temporary Packer security groups created by Amigo. Whilst we wrote a script to delete the unused security groups, we wanted to find a way to stop Amigo creating all of these security groups.

## What is the solution?
We decided to add SSM support to Packer and create two dedicated security groups (one for CODE and another for PROD), so that the security groups do not need any inbound access rules and also so that we no longer create a new temporary security group for each new Packer instance created by Amigo.

> Support for the AWS Systems Manager session manager lets users manage EC2 instances without the need to open inbound ports, or maintain bastion hosts...Once the tunnel has been created all SSH communication will be tunnelled through SSM to the remote instance. https://www.packer.io/docs/builders/amazon/ebs#ssh_interface-1

## How to test
We've tested this on CODE.

When deploying to PROD:
1. upload the new cloudformation to amigo PROD
2. copy the security group id that is created
3. add this to the config in dynamodb
4. push the latest change to teamcity and deploy

No more temporary Packer security groups are created by Amigo, we just have 2 dedicated security groups.

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -25,7 +25,9 @@ Parameters:
   PackerVersion:
     Description: What version of Packer to install
     Type: String
-    Default: 1.1.2
+  PackerInstanceProfile:
+    Description: Instance profile given to instances created by Packer #find this in the PackerUser-PackerRole in IAM
+    Type: String
   TLSCert:
     Type: String
     Description: ARN of a TLS certificate to install on the load balancer
@@ -110,6 +112,10 @@ Resources:
             - s3:GetObject
           Resource:
             - !Sub 'arn:aws:s3:::${AmigoDataBucket}/*'
+        - Effect: Allow
+          Action:
+            - iam:GetInstanceProfile
+          Resource: !Ref PackerInstanceProfile
       Roles:
       - !Ref 'RootRole'
 
@@ -133,6 +139,7 @@ Resources:
           - ssm:ListInstanceAssociations
           - ssm:DescribeInstanceProperties
           - ssm:DescribeDocumentParameters
+          - ssm:StartSession
           - ssmmessages:CreateControlChannel
           - ssmmessages:CreateDataChannel
           - ssmmessages:OpenControlChannel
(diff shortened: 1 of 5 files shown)


cloudformation.yaml
```diff
@@ -267,6 +274,9 @@ Resources:
           unzip -d /opt/packer /tmp/packer_*_linux_amd64.zip
           echo 'export PATH=${!PATH}:/opt/packer' > /etc/profile.d/packer.sh
 
+          wget -P /tmp https://s3.amazonaws.com/session-manager-downloads/plugin/latest/ubuntu_64bit/session-manager-plugin.deb
+          dpkg -i /tmp/session-manager-plugin.deb
```
philmcmahon, line 278: in a follow up PR, we should move both this step and the preceeding step to install packer into the AMI so that it doesn't run on every boot, but I think we can keep it here for now as it's a significant security improvement

philmcmahon (approved):

Nirvikalpa108:
This change is now on PROD and working 👍🏽

cloudformation.yaml
```diff
@@ -110,6 +112,10 @@ Resources:
             - s3:GetObject
           Resource:
             - !Sub 'arn:aws:s3:::${AmigoDataBucket}/*'
+        - Effect: Allow
+          Action:
+            - iam:GetInstanceProfile
+          Resource: !Ref PackerInstanceProfile
```
akash1810, line 118: I'm looking at the policies we define as part of the CDK migration (see #598). I can't quite work out why this policy is needed (other than tracking it down to this PR). Are there any links available that explain it?

philmcmahon: I think we needed to add this so that the amigo instances are able to initial SSM sessions with the packer instances.

PR #527, opened 2021-01-20
Remove enhanced networking patch

sihil:
## What does this change?
This removes the role that provides enhanced networking as it is no longer needed. Post Trusty (14.04) this is installed on Ubuntu AMIs by default (see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html#enhanced-networking-ena-ubuntu).

## How to test

All AMIs should bake and work as per normal.

No interruption to service.

Omitting this should actually reduce risk.


philmcmahon (approved):

adamnfish:
We learned later (today) that removing roles from Amigo also removes them from the UI, so it isn't easy for people to remove this role from their bakes*. Any recipes that still include the removed role will fail.

Ideally we'd be sure to update all existing recipes prior to removing a role, to remove all uses of that role and to confirm that all our recipes still bake successfully.

Clicking "edit" then "save" works because the removed role won't appear on the edit screen, but this isn't obvious or intuitive.

PR #529, opened 2021-01-22
Add role to reserve disk space for .ssh directory

davidfurey:
## What does this change?

On many occasions teams have been unable to SSH to instances using SSM because the instance has run out of disk space.  This adds a role which reserves 10MB of diskspace for `/home/ubuntu/.ssh` which should avoid this issue.

## How to test

Based on https://docs.google.com/document/d/1IvLH8jt5JNYWwsr8S2mDEISVtSDwCnjBMfSmmYun_FI/edit

- [x] Create an AMI with this role
- [x] Create an instance of this AMI
- [x] Use ssm ssh to get onto the instance
- [x] Run `fallocate -l $(($(df . --output=avail | tail -1) * 1024)) filler` to fill the root partition
- [x] Try to use ssm ssh again from another local terminal. It should work

### DIFF
--- a/roles/ssh-reserve-disk-space/README.md
+++ b/roles/ssh-reserve-disk-space/README.md
@@ -1,4 +1,4 @@
 SSH reserve disk space
 ======================
 
-This role reserves disk space for the authorized_keys file so that you can connect to an instance that has run out of disk space.
+This role reserves disk space for amazon ssm and the authorized_keys file so that you can connect to an instance that has run out of disk space.
--- a/roles/ssh-reserve-disk-space/README.md
+++ b/roles/ssh-reserve-disk-space/README.md
@@ -1,0 +1,4 @@
+SSH reserve disk space
+======================
+
+This role reserves disk space for the authorized_keys file so that you can connect to an instance that has run out of disk space.
--- a/roles/ssh-reserve-disk-space/tasks/main.yml
+++ b/roles/ssh-reserve-disk-space/tasks/main.yml
@@ -22,3 +22,27 @@
     owner: ubuntu
     group: ubuntu
     mode: '0700'
+
+- name: Reserve 10MB for ssm
+  command: "dd if=/dev/zero of=/root/ssm.img bs=1024 count=10k"
+  args:
+    creates: "/root/ssm.img"
+
+- name: Create ext2 filesystem
+  filesystem:
+    fstype: ext2
+    dev: /root/ssm.img
+
+- name: Mount ssm
+  mount:
+    path: /var/lib/amazon/ssm
+    src: /root/ssm.img
+    fstype: ext2
+    state: mounted
+
+- name: Set permissions for ssm dir
+  file:
+    path: /var/lib/amazon/ssm
+    owner: root
+    group: root
+    mode: '0700'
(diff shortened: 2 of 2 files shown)


sihil:
Clever idea @davidfurey. I'm not convinced that failure to write to `~/.ssh` is the failure here (although that probably is a path). It's generally the case that you can't SSH to a machine with a full disc, even if you are not writing fresh keys. I think this might be down to sshd being unable to write to /tmp or /var (/var/spool?) but it's been a long time. 

Historically this was guarded against by splitting the disk and isolating /tmp and /var/logs from the rest of the root volume which was often the main offender but this is rarely done or supported in AWS. Also, many file systems (including ext4) reserve 5% or so for use only by the root user which means that root can always log in - but this is of little help when we've disabled the ability for root to log in.

In short, I'm pretty sure this won't work as is but it would be good to try and address the issue more permanently.

sihil:
Oh also - this is failing as the SSM execution is failing so we'd need to grant space such that the SSM agent works.

mbarton:
> We'd need to grant space such that the SSM agent works

I did a ps after I used `ssm ssh` to access the box. Looks like the scripts end up in `/var/lib/amazon/ssm`:

```
root         492       1  0 Jan21 ?        00:00:48 /snap/amazon-ssm-agent/1566/amazon-ssm-agent
root        9434     492  0 14:40 ?        00:00:00 /snap/amazon-ssm-agent/1566/ssm-document-worker dd20e0a4-c533-4553-989c-b41a5893f74f i-02e3ce09d3cddfee4
root        9707     492  0 14:41 ?        00:00:00 /snap/amazon-ssm-agent/1566/ssm-document-worker b631ab37-6762-4339-b66d-0259db0bbb16 i-02e3ce09d3cddfee4
root        9712    9707  0 14:41 ?        00:00:00 sh -c /var/lib/amazon/ssm/i-02e3ce09d3cddfee4/document/orchestration/b631ab37-6762-4339-b66d-0259db0bbb16/awsrunShellScript/0.awsrunShellScript/_script.sh
root        9713    9712  0 14:41 ?        00:00:00 /bin/sh /var/lib/amazon/ssm/i-02e3ce09d3cddfee4/document/orchestration/b631ab37-6762-4339-b66d-0259db0bbb16/awsrunShellScript/0.awsrunShellScript/_script.sh
root        9717     492  5 14:41 ?        00:00:00 /snap/amazon-ssm-agent/1566/ssm-session-worker michael.barton-034fe7b070d4e28cc i-02e3ce09d3cddfee4
ubuntu      9840    9830  0 14:41 pts/0    00:00:00 grep --color=auto ssm
```

so perhaps we could reserve space there? Ideally that wouldn't break in the future, perhaps there's a way of configuring the location? In an ideal world we could install the entire SSM agent on to an isolated partition but I imagine that would be fiddly if it's only available as a snap

davidfurey:
Interesting.  I applied these changes to a code instance, and as you expected it didn't work.  Once I applied the same approach to reserve space for `/var/lib/amazon`, ssm ssh was able to connect even when the root filesystem is full.

sihil:
Might be worth looking on a week old instance to understand what else, if anything, is in `/var/lib/amazon`. Just to ensure we are allocating enough space. It would be sad to replace one problem with another, were we to execute lots of SSM commands in such a way that we exhaust this new partition!

davidfurey:
Good point.  The instance I tested on was 2 weeks old, and /var/lib/amazon is 326KB. 

10MB was quite arbitrary. I kept it low because I originally planned to use a ramdisk, but then worried about the impact OOM issues could have.  Since it is now on disk, there is no particular reason to keep it so low.

mbarton (approved):
LGTM, nice one

PR #530, opened 2021-01-22
Move wazuh agent parameters to ParameterStore

jorgeazevedo:
## What does this change?

The `wazuh-agent` role currently takes 3 input parameters in amigo

 * agent_version
 * manager_ip
 * secret_arn

This becomes painful as more and more recipes use this role, as the parameters needs to be copy pasted around, which also means that if we need to upgrade the agent version, we'll have to make many edits.

This PR re architects this. Those 3 values are now stored in aws ParameterStore

* /stage/stack/amigo-role/wazuh-agent/authentication-secret
* /stage/stack/amigo-role/wazuh-agent/manager-address
* /stage/stack/amigo-role/wazuh-agent/agent-version

With an optional `stage` parameter that can be defined in amigo to override the default value `PROD`. So multiple sets of these 3 parameters might live at one time, which can be useful when testing new agent versions etc

## How to test
Deploy to code and bake `wazuh-agent`. Boot the image in a fresh ec2 instance. Ssh to it and check `journalctl -u wazuh-agent` to see if the agent connected successfully to the manager.

## What then

The recipes that currently use this need to be edited to remove parameters and rebaked/redeployed


philmcmahon (approved):
This looks like a good improvement, and cleans up the role a bit in the process too

katebee:
Ideally the `agent_version` would be kept in sync with the Manager version. However, agents should still report if they are on an earlier version:

> The compatibility between Wazuh agent and Wazuh manager is guaranteed when the Wazuh manager has a newer or equal version than the Wazuh agent.

Source: [Wazuh documentation](https://documentation.wazuh.com/4.0/upgrade-guide/compatibility_matrix/index.html#components-compatibility)

jorgeazevedo:
That's great to know @katebee, it makes for an easier upgrade path!

Some follow up for our meeting yesterday. I've switched to fetching the secret from the infosec, so that's the duplication taken care of!

On the cross account ParameterStore. From what I gather, secrets manager lends itself better for cross account usage, as it takes an arn as an argument which includes the account id.

```
aws secretsmanager get-secret-value --secret-id my::arn::
```

ParameterStore on the other hand, takes a path. Which account this refers to is implicitly defined by what credentials are used

```
aws ssm get-parameter --name /my/path/
```

So going across accounts is possible but it becomes quite impractical. I think we'd have to set up a separate IAM user, call `assume-role` to fetch temporary credentials and inject them into the environment in a way that the aws cli picks them up. (see [this forum post](https://forums.aws.amazon.com/thread.jspa?threadID=269056) and [this blog post](https://aws.amazon.com/premiumsupport/knowledge-center/iam-assume-role-cli/)).

So yeah, this is a bit tricky. What do you reckon?

jorgeazevedo:
@katebee are you OK with us merging it like this? The agent upgrade story isn't ideal, but it'd be easier to change once we get rid of the role parameters in amigo.

katebee (approved):
👍

PR #533, opened 2021-01-29
Parameterise architecture for node role.

philmcmahon:
In order to support both x64 and amr64 architectures for amigo bakes it makes senst to parameterise the architecture in this role. Ideally we'd calculate this based off the architecture of the instance packer is running on, but this is a fairly quick fix in the short term.

**Testing**
This has been tested on CODE https://amigo.code.dev-gutools.co.uk/recipes/devx-arm64-test.

I've also verified that recipes where 'architecture' isn't specified still work

### DIFF
--- a/roles/node/defaults/main.yml
+++ b/roles/node/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 node_version: 4.x
+architecture: x64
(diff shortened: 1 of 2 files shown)


jorgeazevedo (approved):
Yay!

PR #535, opened 2021-02-02
Show a diff of modified packages between bakes

philmcmahon:
## What does this change?
This change makes use of google's [diff-match-patch](https://github.com/google/diff-match-patch) library to generate a diff of the package lists of the current bake and the previous bake. This results in a more useful indication of what changed in the current bake.

## How to test
Perform at least 2 bakes on a recipe with some change between them. You should see the diff show up on the bake page. Locally you can experiment with [this recipe](http://localhost:9000/recipes/node-is-theft/bakes/3)

Next time someone has an issue they suspect is caused by a change in an AMI it should be more obvious to them what has changed. 

## Images

Screenshot 2021-02-02 at 10 19 32

### DIFF
--- a/public/stylesheets/main.css
+++ b/public/stylesheets/main.css
@@ -23,6 +23,10 @@ body { padding-top: 70px; }
     color: #DCDCDC; /* iTerm2's default theme */
 }
 
+.package-list {
+    white-space: nowrap;
+}
+
 .btn--success:after {
     content: 'Successful';
     display: inline-block;
--- a/public/stylesheets/main.css
+++ b/public/stylesheets/main.css
@@ -23,6 +23,20 @@ body { padding-top: 70px; }
     color: #DCDCDC; /* iTerm2's default theme */
 }
 
+.package-diff-insert {
+    color:rgb(239, 246, 252);
+    background: rgb(10, 108, 54);
+    text-decoration: underline;
+}
+
+.package-diff-delete {
+    color: rgb(239, 246, 252);
+    background: rgb(144, 37, 42);
+    text-decoration: line-through;
+}
+
+
+
 .package-list {
     white-space: nowrap;
 }
--- a/public/stylesheets/main.css
+++ b/public/stylesheets/main.css
@@ -13,13 +13,10 @@ body { padding-top: 70px; }
     border-radius: 0 0 0.5em 0.5em;
 }
 
-#packer-output {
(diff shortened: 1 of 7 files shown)


jorgeazevedo:
There's an @ missing from the error message in [app/views/showBake.scala.html](https://github.com/guardian/amigo/pull/535/files#diff-689d377ff6d2700d9fe1e6e6a3a22610899a39be617d48e292ab22086d13b805R56) and there's a layout issue with the placement of the message.

AMIgo_and_amigo_—_-bash_—_207×38_and_AMIgo

I think that message paragraph needs to be in the same column as the "This data comes from the OS default package manager[...]" paragraph. I tried it locally and it seems to fix the issue

AMIgo

jorgeazevedo:
Is this the second message here also intended to be in a well? http://localhost:9000/recipes/df-test/bakes/2

AMIgo

philmcmahon:
> Is this the second message here also intended to be in a well? http://localhost:9000/recipes/df-test/bakes/2

Yep good spot!

jorgeazevedo (approved):
I think this is in a good state for us to deploy and get some feedback! Maybe we can open a ticket to write down some ideas for future tweaks

public/stylesheets/main.css
```diff
@@ -13,14 +13,27 @@ body { padding-top: 70px; }
...
+    overflow-y: scroll;
+    max-height: 400px; /* This gets updated to match the screen size using JS */
+    color: #DCDCDC; /* iTerm2's default theme */
 }
 
-.bake-log {
-    color: #DCDCDC; /* iTerm2's default theme */
+.package-diff-insert {
+    color:rgb(239, 246, 252);
+    background: rgb(10, 108, 54);
```
akash1810, line 25:
```suggestion
    color: #EFF6FC;
    background: #0A6C36;
```

A small change to keep consistent colour declaration styles within the repo.

public/stylesheets/main.css
```diff
@@ -13,14 +13,27 @@ body { padding-top: 70px; }
...
+    text-decoration: underline;
+}
+
+.package-diff-delete {
+    color: rgb(239, 246, 252);
+    background: rgb(144, 37, 42);
```
akash1810, line 31:
```suggestion
    color: #EFF6FC;
    background: #90252A;
```

A small change to keep consistent colour declaration styles within the repo.

PR #537, opened 2021-02-05
Upgrade aws-kinesis-agent to 2.0.1

tomrf1:
Which supports ARM.
The earlier version has a SQLite dependency that breaks on ARM instances


philmcmahon (approved):
Nice one!

PR #543, opened 2021-02-16
Add concurrent bake limit

sihil:
## What does this change?
This implements a brutally simple limiter to cap the number of bakes running at once. It seems that AMIgo can become overwhelmed when a large number of scheduled bakes kick off and this should limit the impact of that.

## How to test
Deploy to CODE and try to kick off more than five bakes. Anything over five should not start until prior bakes have completed.

We have fewer scheduled bake failures.

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -25,7 +25,7 @@ import notification.{ AmiCreatedNotifier, LambdaDistributionBucket, Notification
 import org.joda.time.Duration
 import org.quartz.Scheduler
 import org.quartz.impl.StdSchedulerFactory
-import packer.PackerConfig
+import packer.{ PackerConfig, PackerRunner }
 import play.api.{ BuiltInComponentsFromContext, Configuration, Logger }
 import play.api.ApplicationLoader.Context
 import play.api.i18n.I18nComponents
@@ -185,11 +185,13 @@ class AppComponents(context: Context)
 
   val amigoDataBucket: Option[String] = configuration.getString("amigo.data.bucket")
 
+  val packerRunner = new PackerRunner(configuration.getInt("packer.maxInstances").getOrElse(5))
+
   val scheduler: Scheduler = StdSchedulerFactory.getDefaultScheduler()
 
   val scheduledBakeRunner: ScheduledBakeRunner = {
     val enabled = identity.stage == "PROD" // don't run scheduled bakes on dev machines
-    new ScheduledBakeRunner(identity.stage, enabled, prismAgents, eventBus, ansibleVariables, amiMetadataLookup, amigoDataBucket)
+    new ScheduledBakeRunner(identity.stage, enabled, prismAgents, eventBus, ansibleVariables, amiMetadataLookup, amigoDataBucket, packerRunner)
   }
   val bakeScheduler = new BakeScheduler(scheduler, scheduledBakeRunner)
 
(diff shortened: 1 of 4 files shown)


sihil:
Logs from kicking off some bakes in CODE with this change:
```
Feb 16, 2021 @ 10:27:17.875 AMIgo current bake total: 1, running: 0, starting: 1
Feb 16, 2021 @ 10:27:20.580 AMIgo current bake total: 2, running: 1, starting: 1
Feb 16, 2021 @ 10:27:22.575 AMIgo current bake total: 3, running: 2, starting: 1
Feb 16, 2021 @ 10:27:25.595 AMIgo current bake total: 4, running: 3, starting: 1
Feb 16, 2021 @ 10:27:28.309 AMIgo current bake total: 5, running: 4, starting: 1
Feb 16, 2021 @ 10:27:31.154 AMIgo current bake total: 6, running: 5, starting: 0
Feb 16, 2021 @ 10:29:08.466 AMIgo current bake total: 7, running: 5, starting: 0
Feb 16, 2021 @ 10:36:27.302 AMIgo current bake total: 6, running: 4, starting: 1
Feb 16, 2021 @ 10:36:42.496 AMIgo current bake total: 5, running: 4, starting: 1
Feb 16, 2021 @ 10:36:42.999 AMIgo current bake total: 4, running: 4, starting: 0
Feb 16, 2021 @ 10:40:44.120 AMIgo current bake total: 3, running: 3, starting: 0
Feb 16, 2021 @ 10:40:47.986 AMIgo current bake total: 2, running: 2, starting: 0
Feb 16, 2021 @ 10:41:40.175 AMIgo current bake total: 1, running: 1, starting: 0
Feb 16, 2021 @ 10:43:49.419 AMIgo current bake total: 0, running: 0, starting: 0
```

app/components/AppComponents.scala
```diff
@@ -185,11 +185,13 @@ class AppComponents(context: Context)
 
   val amigoDataBucket: Option[String] = configuration.getString("amigo.data.bucket")
 
+  val packerRunner = new PackerRunner(configuration.getInt("packer.maxInstances").getOrElse(5))
```
philmcmahon, line 188: what's the motivation for having this as a class variable rather than keeping PackerRunner as an object and using the `packerConfig` property to get the value?

sihil: I hadn't noticed the packer config, but it doesn't make any sense for `maxInstances` to change between calls to the methods on the instance of the class (or object if you retained that). If the parameter varied between calls then it would lead to a lot of potentially weird behaviour in the number of tasks that were kicked off either immediately or when the task ends. You'd probably have to guard against negative values being passed into `take` as well for example (maybe it deals with that). I don't really want to think about the kind of bugs it makes possible.

Given that `PackerConfig` is constant that could be refactored into a class parameter too.

philmcmahon: 👍🏻 I was just coming from perspective of 'classes bad objects good'

philmcmahon (approved):

app/components/AppComponents.scala
sihil: I think functions on objects are good for containing pure code. This code isn't pure anyway it has side effects of starting external processes and threads to monitor them. We're already going to hell for this before turning it into a class.

app/packer/PackerRunner.scala
```diff
@@ -24,13 +25,13 @@ object PackerRunner extends Loggable {
...
    */
   def createImage(stage: String, bake: Bake, prism: PrismAgents, eventBus: EventBus, ansibleVars: Map[String, String], debug: Boolean, amiMetadataLookup: AmiMetadataLookup, amigoDataBucket: Option[String])(implicit packerConfig: PackerConfig): Future[Int] = {
-    val playbookYaml = PlaybookGenerator.generatePlaybook(bake.recipe, ansibleVars)
-    val playbookFile = Files.createTempFile(s"amigo-ansible-${bake.recipe.id.value}", ".yml")
-    Files.write(playbookFile, playbookYaml.getBytes(StandardCharsets.UTF_8)) // TODO error handling
-
     val sourceAmi = bake.recipe.baseImage.amiId.value
     val amiMetadata = amiMetadataLookup.lookupMetadataFor(sourceAmi).right.getOrElse(throw new IllegalStateException(s"Unable to identify the architecture for $sourceAmi"))
 
+    val playbookYaml = PlaybookGenerator.generatePlaybook(bake.recipe, ansibleVars + ("arch" -> amiMetadata.architecture))
```
sihil, line 31: oops, this change snuck in. oh well. out now.

PR #545, opened 2021-02-18
Add retry logic to friendly-hostnames role

davidfurey:
In the past 2 days, the Ophan team have seen this error from the friendly-hostnames role:

```
An error occurred (InvalidInstanceID.NotFound) when calling the CreateTags operation: The instance ID 'i-059607dad87f615d7' does not exist
```

We are quite surprised that this is happening, since we had assumed that the aws-await-tags step would ensure that the EC2 API was ready, but it seems like the [apis eventual consistency](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#eventual-consistency) isn't as straightforward as that.

Rather than spend too much more time investigating what the rules are of the AWS API, I've added retry logic to the friendly-hostnames role.  This is copied from the aws-await-tags role.


philmcmahon (approved):
LGTM 👍🏻

rtyley (approved):
Looks good to me!

PR #548, opened 2021-02-25
Add JDK 11 as a role

sihil:
## What does this change?
JRE11 has been supported by Scala for [some time now](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#version-compatibility-table) and has some [key optimisations for ARM64 machines](https://bugs.openjdk.java.net/browse/JDK-8189104).

This adds a role that should be a simple swap out for recipes switching from version 8.

## How to test
This impacts baked AMIs so best tested by converting an AMI over.

### DIFF
--- a/roles/java11/meta/main.yml
+++ b/roles/java11/meta/main.yml
@@ -1,0 +1,4 @@
+---
+dependencies:
+  - role: apt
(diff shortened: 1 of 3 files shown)


philmcmahon (approved):
🚀

PR #551, opened 2021-02-26
Move Kong base image to bionic LTS. Version postgres client to match

tonytw1:
## What does this change?

Updates Kong role to assume an ubuntu 18.04 base image rather than 16.04.

Provoked by the AWS CLI Python 3.6 fail this week but also gets us ready for the 16.04 end of support.

Levels the postgres client from 9.5 to the 10 provided in bionic (this is safe because it's only used for ad hoc queries).

## How to test

Baked by deploying this branch to CODE.
Deployed the AMI to CODE Kong.
Checked access logs and logstashing.

No used facing impact should be visible.

Wierd Kong / Perl errors.
Should be safe as Kong provide a specific bionic .deb package.

### DIFF
--- a/roles/kong/defaults/main.yml
+++ b/roles/kong/defaults/main.yml
@@ -1,4 +1,4 @@
 ---
 kong_version: 0.14.0
 kong_user: content-api
-postgres_version: 9.5
+postgres_version: 10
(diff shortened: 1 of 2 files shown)


Fweddi (approved):
LGTM!

PR #553, opened 2021-03-01
Prevent AMIgo housekeeping from removing TeamCity AMIs

jacobwinch:
## What does this change?

Last month we experienced a problem launching new TeamCity agents, because [AMIgo had deleted the AMI associated with our TeamCity Cloud Profile](https://docs.google.com/document/d/1Qu5tPLMcAFzHx9Tzu3U_aAKHiEZkvdQKgpsGKrXpb9Y/edit).

In order to avoid a repeat of this problem, I was hoping to update the AMI associated with our Cloud Profile. Unfortunately the [bakes for this recipe](https://amigo.gutools.co.uk/recipes/teamcity-agent) have been failing for several weeks, so as things stand we run the risk of losing our only working AMI.

I'd like to prevent AMIgo from deleting `teamcity-agent` AMIs as a short-term measure so that we have time to debug and fix the failing bakes.

## How to test

I've added a unit test to cover the new behaviour.

* Our [current `teamcity-agent` AMI](https://amigo.gutools.co.uk/recipes/teamcity-agent/bakes/147) should not be deleted by AMIgo's housekeeping
* Old, unused AMIs associated with other recipes should still be deleted

* If we forget to revert this PR we will accumulate a significant number of `teamcity-agent` AMIs over time

### DIFF
--- a/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
+++ b/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
@@ -23,7 +23,7 @@ object MarkOldUnusedBakesForDeletion {
       duration.getStandardDays > MAX_AGE
     }
 
-    // Exclude TeamCity Agent AMIs, which have been incorrectly assumed to be as unused in the past. This happens
+    // Exclude TeamCity Agent AMIs, which have been incorrectly assumed to be unused in the past. This happens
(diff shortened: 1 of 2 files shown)


philmcmahon (approved):
Sensible workaround 👍🏻

PR #540, opened 2021-02-12
Update aws-cloudwatch-agent role to support ARM64 architecture

Mullefa:
## What does this change?

\* worked on with @Fweddi 

Update the `aws-cloudwatch-agent` role to dynamically install the `amazon-cloudwatch-agent.deb` dynamically based on the architecture (architectures supported: amd64, arm64).

Uses the Ansible fact `ansible_architecture` to determine the architecture.

Tested on CODE by:
1. creating a new base image `ARM ubuntu-xenial` (as was done on PROD)
2. cloning the recipe `ubuntu-xenial-capi` (clone name: `ubuntu-xenial-capi-ARM`) and changing the base image to `ARM ubuntu-xenial`
3. baking recipes [`ubuntu-xenial-capi`](https://amigo.code.dev-gutools.co.uk/recipes/ubuntu-xenial-capi/bakes/2) and [`ubuntu-xenial-capi-ARM`](https://amigo.code.dev-gutools.co.uk/recipes/ubuntu-xenial-capi-ARM/bakes/1) and eyeballing the logs to check the appropriate package was installed

PR #555, opened 2021-03-02
update README for node role, to include 'architecture' param

twrichards:
The architecture param was introduced in https://github.com/guardian/amigo/pull/533 but not documented, so adding it here for future travelers.

### DIFF
--- a/roles/node/README.md
+++ b/roles/node/README.md
@@ -12,6 +12,8 @@ Using this approach the binary will be unpacked to `/usr/local/node` and a `node
 
 * `packages` - npm packages to install globally. e.g. `packages: [pm2, reqwest]`
 
+* `architecture` - defaults to `x64` (but for example, one can specify `arm64` for graviton instances)
+
 Using this approach, node binaries (node, npm, npx) will be accessible at `/usr/local/node/`. The `node` binary is linked to in `/usr/bin` so should be available via the `PATH` environment variable.
 
 #### DEPRECATED Approach


twrichards:
FYI @rBangay

tomrf1 (approved):
👍

Fweddi:
Guy and I encountered this scenario with another role, where we made the 'architecture' injection a case of automatic detection rather than a manual parameter (See https://github.com/guardian/amigo/pull/540/files, and https://docs.google.com/document/d/1J6zBuz9P5E57XFNKj2Pb75Z1AYDb01bzz6_FFKtQo50/edit#heading=h.ypqm5kuc0fv5 for more detail). Might be worth adding here?

twrichards:
> Guy and I encountered this scenario with another role, where we made the 'architecture' injection a case of automatic detection rather than a manual parameter (See https://github.com/guardian/amigo/pull/540/files, and https://docs.google.com/document/d/1J6zBuz9P5E57XFNKj2Pb75Z1AYDb01bzz6_FFKtQo50/edit#heading=h.ypqm5kuc0fv5 for more detail). Might be worth adding here?

Excellent suggestion. Phil's original PR mentions that in fact #533 - I'm fairly rusty on ansible but I suspect we could leave this `architecture` param but update the default to be calculated from `ansible_architecture`, I'll take a punt at it later unless @philmcmahon beats me to it

PR #556, opened 2021-03-03
Revert "Prevent AMIgo housekeeping from removing TeamCity AMIs"

jacobwinch:
Reverts guardian/amigo#553.

https://github.com/guardian/amigo/pull/554 fixes baking for the [`teamcity-agent` recipe](https://amigo.gutools.co.uk/recipes/teamcity-agent) and I have updated all TeamCity agents to use a new AMI. 

Consequently, I think it's safe to revert this (as long as we remember to regularly update the AMI associated with our Cloud Profile going forwards). Alternatively, we could leave this hack in until we have a proper (i.e. automated) solution for updating the TeamCity agent AMI. What do you think @philmcmahon?

### DIFF
--- a/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
+++ b/app/housekeeping/MarkOldUnusedBakesForDeletion.scala
@@ -22,17 +22,10 @@ object MarkOldUnusedBakesForDeletion {
       val duration = new Duration(bake.startedAt, now)
       duration.getStandardDays > MAX_AGE
     }
-
-    // Exclude TeamCity Agent AMIs, which have been incorrectly assumed to be unused in the past. This happens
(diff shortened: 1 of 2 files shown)


philmcmahon:
Ah I'm not sure! I don't think we'll remember to keep updating the AMI in the cloud profile so it might be best to keep this in until we can think of a better solution - or just accept that sometimes this issue will break teamcity.

jacobwinch:
> I don't think we'll remember to keep updating the AMI in the cloud profile so it might be best to keep this in until we can think of a better solution

OK, let's leave the hack in for now! I will leave this PR open too; I think it makes it slightly less likely that we'll completely forget about this problem.

PR #558, opened 2021-03-12
Add Fluentbit role

nicl:
Fluentbit is a small and fast log shipper. It can send logs to Kinesis (our most common destination) out of the box.

Full docs here:

https://docs.fluentbit.io/manual/

For initial motivation see: https://github.com/guardian/platform/pull/1438.


roles/fluentbit/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+ubuntu_version: bionic
```
jfsoul, line 2: my ansible is a bit rusty but semantically should this be in a `defaults` directory rather than vars?

jfsoul (approved):

roles/fluentbit/vars/main.yml
nicl: Ah yes. So both work, but there are differences in priority and defaults is much clearer in any case.

mchv:
Nice one @nicl  💎

PR #563, opened 2021-04-08
Install arm64 version of docker when on arm64 architecture.

philmcmahon:
This does what is says on the tin - sets up the docker role to install the correct version of docker-ce based off the current architecture.

How to test
========
Deploy to Amigo CODE, run the teamcity-agent recipe (or any recipe that uses docker): amigo.code.dev-gutools.co.uk/


jorgeazevedo (approved):
Looks solid to me!

PR #565, opened 2021-04-15
Add end of life date to base images 

philmcmahon:
## What does this change?
This PR aims to help prepare us for ubuntu Xenial being deprecated at the end of this month. It adds warning colours and labels next to Recipes and AMI's that are using a base image which is due to expire soon.

To make this work I've added a new property to the `BaseImage` case class and associated forms  - eolDate. Play Forms provides an adequate date picker for this purpose, it could be prettier but does  the job.

Most of the rest of the changes are to the views in order to expose this information. The CSS isn't super pretty (I copied it from riffraff) but given that our aim is to have zero images that are EOL soon I think it's probably fine to look a bit ugly.

Once this change is live we'll need to manually go through the existing base images and add an eolDate to them all. A link is provided in the UI to the ubuntu releases page to help with this. 

The different views where the eolDate is exposed look like this:
Screenshot 2021-04-14 at 18 42 09
Screenshot 2021-04-14 at 18 17 41
Screenshot 2021-04-14 at 18 17 33

Finally, this change also makes recipe usages information (i.e. whether a recipe is used by any instances or launch configs) available in the 'show bake' view - so you can see at a glance whether a base image is in use anywhere:

Screenshot 2021-04-15 at 10 36 16

## How to test
This is currently live on CODE: amigo.code.dev-gutools.co.uk/

### DIFF
--- a/app/views/fragments/usagesColumn.scala.html
+++ b/app/views/fragments/usagesColumn.scala.html
@@ -1,0 +1,15 @@
+@(
+    recipe: Recipe,
+    usages: Map[Recipe, prism.RecipeUsage]
+)
+
+@defining(usages.getOrElse(recipe, prism.RecipeUsage.noUsage)) { usage =>
+    <td class="recipe-usage" title="Used in @usage.instances.size instance@if(usage.instances.size != 1) {s} and @usage.launchConfigurations.size launch configuration@if(usage.launchConfigurations.size != 1) {s}">
+        <a href="@routes.RecipeController.showUsages(recipe.id)">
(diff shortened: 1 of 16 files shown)


app/views/fragments/usagesColumn.scala.html
```diff
@@ -0,0 +1,15 @@
+@(
```
philmcmahon, line 1: this is old code pulled out of recipes.scala.html

jorgeazevedo (approved):
Nice work! Adding an end of life property to the base image is a solid approach

PR #567, opened 2021-04-20
Add support for single instance configs

jorgeazevedo:
## What does this change?
This makes the wazuh registration script try to fetch tags from the instance if the result from the ASG fetch is an empty string or the word "None". This will cover single instances, but ends up also covering if an instance only has permission to query its own tags and now its ASG ones.

## How to test
Writing an if statement in bash is a known high risk situation. I've actually tested this code by running it
- In a single instance
- In an instance in an ASG without asg:describeTags, but with ec2:describeTags
- In an instance in an ASG with asg:describeTags

For completeness' sake, it would be worth baking an image in amigo CODE and then launching it somewhere to see it running end to end. I'll do that before merging for sure

The single instance that cannot currently connect to wazuh will indeed connect

There's over 500 servers relying on this code. If this is broken, next time they deploy the wazuh agent will fail to connect

### DIFF
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -87,6 +87,17 @@
         STACK=$(aws autoscaling $ARGS --query "AutoScalingGroups[0].Tags[?Key == 'Stack'].Value" || true)
         STAGE=$(aws autoscaling $ARGS --query "AutoScalingGroups[0].Tags[?Key == 'Stage'].Value" || true)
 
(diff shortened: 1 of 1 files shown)


philmcmahon (approved):
Nice!! 🚀

jorgeazevedo:
testing in CODE worked fine! I booted a single instance without an asg and the script fell back to querying the instances' tags instead. I'll probably merge this on monday and immediately try one of our services to see if it works properly

PR #569, opened 2021-04-27
Fix reading of friendly hostnames file for Bionic

frederickobrien:
[This role](https://amigo.gutools.co.uk/roles#friendly-hostnames) failed when trying to generate a new id using the hostnames.txt file, giving the following error on Ubuntu Bionic:

```
iconv: illegal input sequence at position 0
```

Ubuntu Bionic has iconv version 2.27, while Xenial is 2.23. Bionic has a version of iconv that is seemingly stricter with invalid character encoding so this fix skips over invalid ones:

```
Output control:
  -c                         omit invalid characters from output
```

### DIFF
--- a/roles/friendly-hostnames/files/set-hostname.sh
+++ b/roles/friendly-hostnames/files/set-hostname.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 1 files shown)


rtyley (approved):
We found this while updating our ophan-elasticsearch recipe to Ubuntu Bionic.

philmcmahon:
Thanks @frederickobrien !

PR #571, opened 2021-04-28
Don't try and set plugin alternatives for java

philmcmahon:
## What does this change?
This modifies the teamcity agent role so that it doesn't use update-java-alternatives, as this fails  with the error below

```
sudo update-java-alternatives --set /usr/lib/jvm/java-1.8.0-openjdk-amd64
update-alternatives: error: no alternatives for mozilla-javaplugin.so
update-java-alternatives: plugin alternative does not exist: /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/amd64/IcedTeaPlugin.so
```

## How to test
Run an amigo bake on code then cmd+f the build log for `update-alternatives`

Java 8 becomes default  version on teamcity AMIs.

### DIFF
--- a/roles/teamcity-agent/tasks/main.yml
+++ b/roles/teamcity-agent/tasks/main.yml
@@ -96,7 +96,7 @@
   shell: ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
 
 - name: set default java version # to get available versions run `update-java-alternatives --list`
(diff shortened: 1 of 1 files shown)


sihil (approved):

philmcmahon:
I have confirmed that java/javac -version returns the correct value on a test agent

PR #572, opened 2021-04-29
Show unused recipes separately

philmcmahon:
## What does this change?
This changes the recipes and baseimageussage lists so that they have 2 sections - one for recipes in use (with an active instance or launch configuration), and one for unused recipes.

As part of this work I added a new `getUsagesMap` function which simplifies somewhat the calls  to RecipesUsage from the Recipes/BaseImage controllers.

This should make it easier to see which end of life recipes we need to worry about, and also easier to clean out unused recipes.

## How to test
I've tested this on CODE - it should just be a UI change. See screenshots below

### Recipes view
Screenshot 2021-04-29 at 12 32 42

### Base image view
Screenshot 2021-04-29 at 11 41 29


jorgeazevedo (approved):
This is really useful!!

PR #583, opened 2021-05-07
Add new role to disable systemd dns.

philmcmahon:
## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/roles/teamcity-agent/tasks/main.yml
+++ b/roles/teamcity-agent/tasks/main.yml
@@ -91,8 +91,7 @@
     name: teamcityagent
     enabled: yes
 
-# this fixes an issue with running docker on ubuntu bionic, see here for details: https://github.com/moby/libnetwork/issues/2187
-- name: disable systemd as default ddns server
+- name: disable systemd as default dns server
   shell: ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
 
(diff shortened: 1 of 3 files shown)


roles/teamcity-agent/tasks/main.yml
```diff
@@ -91,8 +91,7 @@
     name: teamcityagent
     enabled: yes
 
-# this fixes an issue with running docker on ubuntu bionic, see here for details: https://github.com/moby/libnetwork/issues/2187
-- name: disable systemd as default ddns server
+- name: disable systemd as default dns server
   shell: ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
```
sihil, line 95: Should this be removed from the `teamcity-agent` role and a dependency added instead?

PR #584, opened 2021-05-13
Add delete bake button

philmcmahon:
## What does this change?
This adds a button to the 'show bake' view of amigo to delete a bake and all the AMIs associated with it. The delete button is only visible where an AMI is not currently in use.

Screenshot 2021-05-13 at 14 12 43

The motivation here is to allow developers to quickly 'roll back' a bake that doesn't work rather than having to undo the change and wait for a new bake. For example, this will significantly simplify the 'fast mode' of the [xenial migration docs](https://docs.google.com/document/d/1igQx9KMCHkYQ3IHGzNJRbuG2HcD6nIOUlJcAn7BFjLE/edit#)

As part of this work I've addded some logic to calculate whether or not a bake is currently being used. This checks not only the base AMI but also any encrypted copies. This logic has been used to add a new 'in use' column to the bake list:

Screenshot 2021-05-13 at 14 10 49

## How to test
I have tested this on AMIgo CODE. There may be some issues with the UI that weren't obvious on CODE (as none of the CODE AMIs are in use) but it should be ok!


app/views/showRecipe.scala.html
```diff
@@ -127,17 +130,24 @@ <h4>Inherited from base image @recipe.baseImage.id</h4>
               </td>
               <td class="absolute-container has-block-link">
                 <button class="btn btn-primary btn-xs absolute-right copy-button" title="Copy to clipboard" data-clipboard-text="@bake.amiId">
-                  <img src="@routes.Assets.versioned("images/clippy.svg")" width="13" alt="Copy to clipboard">
+                  <img src="@routes.Assets.versioned("images/clippy.svg")" width="13" alt="Copy to clipboard"  title="Copy to clipboard">
                 </button>
               </td>
               <td>
-                <a href="https://amiable.gutools.co.uk/ami?imageId=@bake.amiId" target="_blank" rel="noopener noreferrer">View in AMIable</a>
+              @fragments.booleanIcon(RecipeUsage.amiIsUsed(usage, bake.amiId ))
```
sihil, line 137: In the image I find this to be a confusing UI choice. The heading is "In use". It's not super clear if being in use is a good or a bad thing. 

Screenshot 2021-05-13 at 14 10 49

In the sample image we have empty columns and crosses. Does the cross mean it's in use? If there was also a check mark in the column it would perhaps be clearer that that the cross is false/"not in use" and the check mark is true/"in use". That's not the case with crosses in isolation, especially when they don't appear for failed bakes.

I'd perhaps go for a Usages column with a number, possibly even reuse the instances/config thing.

public/images/trash.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="#ffffff" class="bi bi-trash" viewBox="0 0 16 16">
+  <path d="M5.5 5.5A.5.5 0 0 1 6 6v6a.5.5 0 0 1-1 0V6a.5.5 0 0 1 .5-.5zm2.5 0a.5.5 0 0 1 .5.5v6a.5.5 0 0 1-1 0V6a.5.5 0 0 1 .5-.5zm3 .5a.5.5 0 0 0-1 0v6a.5.5 0 0 0 1 0V6z"/>
+  <path fill-rule="evenodd" d="M14.5 3a1 1 0 0 1-1 1H13v9a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2V4h-.5a1 1 0 0 1-1-1V2a1 1 0 0 1 1-1H6a1 1 0 0 1 1-1h2a1 1 0 0 1 1 1h3.5a1 1 0 0 1 1 1v1zM4.118 4 4 4.059V13a1 1 0 0 0 1 1h6a1 1 0 0 0 1-1V4.059L11.882 4H4.118zM2.5 3V2h11v1h-11z"/>
+</svg>
```
sihil, line 4: is this used anywhere?

sihil (commented):
Mainly good but I have a concern around the "In use" UI.

public/images/trash.svg
philmcmahon: nope!

app/views/showRecipe.scala.html
philmcmahon: All good points! Just being consistent with elsewhere seems to make much better sense, this is what it looks like now:

Screenshot 2021-05-17 at 16 58 22

app/views/showBake.scala.html
```diff
@@ -26,9 +29,22 @@ <h1><a href="@routes.RecipeController.showRecipe(bake.recipe.id)">@bake.recipe.i
...
+        <span>@if(inUse){Bake in use}else{Bake is not currently used}</span>
+      </li>
     </ul>
   </div>
 
+  @if(!inUse) {
+    <div class="panel panel-default">
+      <div class="panel-heading">Actions</div>
+      <div class="panel-body">
+        <a href="@routes.BakeController.deleteConfirm(bake.bakeId.recipeId, bake.bakeId.buildNumber)" class="btn btn-danger">Delete bake and all associated AMIs...</a>
```
sihil, line 43: Fairly minor comment - I generally prefer the button to exist but be disabled so that the functionality is discoverable even when viewing bakes that are in use.

app/views/showRecipe.scala.html
```diff
@@ -77,6 +78,8 @@ <h1>@recipe.id.value</h1>
     <div class="panel-body">
     @fragments.customisedRoles(Roles.customisedTransitiveDependency(allRoles, recipe.roles))
       <h4>Inherited from base image @recipe.baseImage.id</h4>
+     <p> <i class="check-square"></i>
+      <i class="unlock"></i>asddadds</p>
```
sihil, line 82: `asddadds` ???

public/stylesheets/main.css
```diff
@@ -110,4 +110,14 @@ td.recipe-usage {
...
+}
+
+.check-icon {
+    padding: 5px;
+    color: green;
+}
+
+.cross-icon{
+    padding: 5px;
+    color: red;
```
sihil, line 122: Left over from the previous version?

sihil (commented):
Few more fairly minor comments.

public/stylesheets/main.css
philmcmahon: thanks

app/views/showRecipe.scala.html
philmcmahon: oops!!!

sihil (approved):

app/views/showBake.scala.html
philmcmahon: good shout

PR #585, opened 2021-05-14
Notify upon bake failure

philmcmahon:
## What does this change?
This PR wires AMIgo up to [Anghammarad](https://github.com/guardian/anghammarad) so that it can send email notifications when a bake fails. Right now this feature relies on the 'request encrypted copy'  feature to identify who owns a recipe - it will only send notifications to the owners of AWS accounts to which an encrypted copy has been requested.

For example, in the screenshot below failure notifications woul be sent to account  095

Screenshot 2021-05-14 at 14 56 42

## How to test
Tested on CODE. This is a backend only change. An example of the output notification can be seen below:

Screenshot 2021-05-14 at 15 04 06

### DIFF
--- a/app/housekeeping/utils/BakesRepo.scala
+++ b/app/housekeeping/utils/BakesRepo.scala
@@ -1,11 +1,13 @@
 package housekeeping.utils
 
 import data.{ Bakes, Dynamo }
-import models.{ Bake, BakeId, BakeStatus }
+import models.{ Bake, BakeId, BakeStatus, NotificationConfig }
+
+import scala.concurrent.ExecutionContext
 
 // Wrapper around dynamo access.
 // Injecting this functionality as a class facilitates unit testing.
-class BakesRepo(implicit dynamo: Dynamo) {
+class BakesRepo(notificationConfig: Option[NotificationConfig])(implicit dynamo: Dynamo, exec: ExecutionContext) {
 
   def getBakes: List[Bake.DbModel] = Bakes.scanForAll()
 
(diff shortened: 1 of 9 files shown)


app/notification/BakeFailedNotifier.scala
```diff
@@ -0,0 +1,73 @@
...
+      notification <- makeNotification(bake, bakeStatus, config)
+    } yield {
+      Anghammarad.notify(notification, config.snsTopicArn, config.snsClient).onComplete {
+        case Success(value) =>
+          log.info(s"Sent notification to anghammarard about recipe ${bake.recipe.id} bake failure. Status: $value")
+        case Failure(exception) =>
+          log.error(s"Failed to send notification about recipe ${bake.recipe.id} bake failure", exception)
+      }
+    }
+    logFailure(notificationResult.isEmpty, notificationConfig.isEmpty, bakeId)
```
jorgeazevedo, line 71: I was a bit confused here, as it reads like a failure will always be logged at the end of `notifyBakeFailed` regardless of outcome. Perhaps bring `logFailure`'s `if (!notificationSuccess) {` conditional to `notifyBakeFailed`?

app/housekeeping/utils/BakesRepo.scala
```diff
@@ -14,7 +16,7 @@ class BakesRepo(implicit dynamo: Dynamo) {
     // Therefore, look up the bake first and check its status.
     Bakes.findById(bakeId.recipeId, bakeId.buildNumber).foreach { bake =>
       if (bake.status == BakeStatus.Running) {
-        Bakes.updateStatus(bakeId, BakeStatus.TimedOut)
+        Bake.updateStatus(bakeId, BakeStatus.TimedOut, notificationConfig)
```
jorgeazevedo, line 19: When I first read this diff I thought the code for `updateStatus` had been moved from `Bakes` to `Bake` which seemed weird. Why add an argument and also move the method to another place? I've read the code for `Bakes.updateStatus` now, so I understand that it deals exclusively with Table updates so it wouldn't really make sense to add a notification concern to it.

Consider perhaps renaming `Bake.updateStatus` to make it explicit that it does a different job from `Bakes.updateStatus`. I don't have a decent suggestion here, `updateTheDatabaseAndAlsoMaybeSendANotificationDepending` doesn't exactly roll off the tongue 😂

jorgeazevedo (approved):
I think this looks solid, just left some optional suggestions to improve readability.

PR #587, opened 2021-06-02
Add AMI ID label to wazuh agent role.

philmcmahon:
## What does this change?
This sets up the wazuh agent role so that when registering an agent with the wazuh server it includes the AMI ID as a [wazuh label](https://documentation.wazuh.com/current/user-manual/capabilities/labels.html)

## How to test
- [x] Deploy this change to AMIgo CODE, bake an AMI and verify that the bake completes. 
- [ ] Then deploy to AMIgo PROD, bake an AMI and redeploy an app using that AMI. 
- [ ] Verify that the app boots succesfully, registers with wazuh and that the AMI ID appears in the wazuh console.

We will be able to aggregate vulnerabilities by ami id

My bash could be dodgy and break the wazuh agent role script 😱  but it's all copy pasted so 🤞🏻


jorgeazevedo (approved):
Thanks @philmcmahon ! Tested this in code

PR #591, opened 2021-06-15
chore: Remove unused Pluto role

akash1810:
## What does this change?

This role is not used by any recipe and can be deleted.

I think this role was created when Editorial Tools were spiking running Pluto in AWS. This was a few years ago, I think the conclusion was not to do it.

## How to test

n/a


akash1810:
cc @guardian/digital-cms who cannot be marked as reviewers for some reason...

jacobwinch (approved):

PR #592, opened 2021-06-16
refactor: Do not download from S3 unnecessarily

akash1810:
## What does this change?

We're currently downloading the same content from S3 into two locations.

This change removes the second download, replacing it with a local disk copy command. This results in less network IO and a faster AMI bake. Logs from previous bakes suggest it takes ~3 minutes to perform the S3 download.

It should also mean a reduction in S3 spend, however looking at the billing dashboard, the savings will be marginal as current spend is in the $10s.

## How to test

- Deploy to CODE
- Bake a recipe that uses the `teamcity-ivy-cache` role, it should complete successfully

A shorter bake time?

## Images

Here's the last three months spend on S3. This change won't save massive amounts of money.

image

### DIFF
--- a/roles/teamcity-ivy-cache/tasks/main.yml
+++ b/roles/teamcity-ivy-cache/tasks/main.yml
@@ -15,8 +15,8 @@
   shell: aws s3 cp s3://teamcity-sbt-cache/ /opt/teamcity/.sbt/ --recursive --quiet
 - name: initial download ivy
   shell: aws s3 cp s3://teamcity-ivy-cache/ /opt/teamcity/.ivy2/cache/ --recursive --quiet
-- name:  download ivy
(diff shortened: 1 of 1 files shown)


jacobwinch (approved):
This seems worthwhile for the performance improvement alone (any cost savings are a bonus!) 👍

akash1810:
Hmm, this role doesn't appear to be used on PROD. I wonder if we can remove it instead?

akash1810:
The `teamcity-agent` role also [does the S3 download twice](https://github.com/guardian/amigo/blob/aca0b0d4332f6c950b8eb7ee9a052b127c506772/roles/teamcity-agent/files/teamcityagent.service#L9-L11), so will make a change to that role in a different PR.

PR #596, opened 2021-06-17
refactor: Use Parameter Store for dist bucket

akash1810:
## What does this change?

Our current best practice is to retrieve the distribution bucket from Parameter Store.

This change updates the main template, adding a new parameter with a default value of the standard location. It also adds a comment to a somewhat confusing policy for future reference.

This is part of the preparation work to move this stack definition into CDK.

See:
  - https://github.com/guardian/cdk/blob/main/docs/005-default-parameter-store-locations.md

### DIFF
--- a/cloudformation.yaml
+++ b/cloudformation.yaml
@@ -118,11 +118,14 @@ Resources:
           Action:
             - sns:Publish
           Resource: !Ref AnghammaradTopicArn
+
+        # Allow us to allow other accounts to retrieve the ImageCopier lambda artifact
         - Effect: Allow
           Action:
           - s3:GetBucketPolicy
           - s3:PutBucketPolicy
           Resource: !Join [ "", [ "arn:aws:s3::*:", !Ref DistributionBucketName ] ]
+
         - Effect: Allow
           Action:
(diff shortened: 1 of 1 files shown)


cloudformation.yaml
```diff
@@ -67,7 +71,7 @@ Resources:
           Action:
           - s3:GetObject
           Resource:
-          - arn:aws:s3::*:deploy-tools-dist/*
+          - !Join [ "", [ "arn:aws:s3::*:", !Ref DistributionBucketName, "/*" ] ]
```
akash1810, line 74: I don't think the first `*` is needed. Indeed [GuCDK](https://github.com/guardian/cdk/blob/04bb5fa0c418f0c2cfc48f05391601f9c0605b3c/src/constructs/iam/policies/s3-get-object.test.ts#L77-L122) will use `:::` rather than `::*:`.

GuCDK also uses a [tighter policy](https://github.com/guardian/cdk/pull/259). I'll move to this format in a later PR to keep this one as close to a no-op as possible.

jacobwinch (approved):

PR #598, opened 2021-06-17
feat: Introduce GuCDK to AMIgo

akash1810:
## What does this change?

In this change we begin the migration from defining the AWS infrastructure in Cloudformation YAML into CDK using the [GuCDK](https://github.com/guardian/cdk) library.

This change does the bare minimum to get us into a place where CD is using a CDK defined stack:
  - Creates a new CDK project with all the config, linting, testing etc.
  - Uses `CfnInclude` to bring the YAML file into the stack
  - Updates the CD step to start using the CDK defined version of the AWS stack

This is [stage one](https://docs.google.com/document/d/17M6TK4uVcMiZAHaTFSo-zahIqrA5eGaKNYU-j4BVUB0/edit) of our migration process and is inspired by AWS's [recommendations](https://www.youtube.com/watch?v=bTC8XV5aLTo).

This change _does_ yield a Cloudformation change (see change set from CODE below), however it is _only_ tag based and doesn't replace any resources. GuCDK [adds two additional tags](https://github.com/guardian/cdk/blob/2e8fc79ed4036446f5414dee7797a6e4f3936268/src/constructs/core/stack.ts#L123-L128) to all resources:
  1. `gu:cdk:version` to track which version of the library is being used
  2. `gu:repo` to help identify which repository a stack is defined in

The change will also slightly change the `Stage` parameter of the stack, moving to a [`GuCDK` parameter](https://github.com/guardian/cdk/blob/e5418a8194c58f9a03c7e191d1fb4f3d2f01aa54/src/constructs/core/parameters/identity.ts#L5-L13) which sets the `allowedValues` property to `["CODE", "PROD"]`.

That is, merging this and having RiffRaff update the stack as part of CD is safe 👍.

Indeed, this branch has been deployed to CODE already w/out issue 🎉.

<details>
<summary>Cloudformation change set JSON from CODE stack</summary>

```json
[
  {
    "resourceChange": {
      "logicalResourceId": "AmigoDataBucket",
      "action": "Modify",
      "physicalResourceId": "amigo-data-code",
      "resourceType": "AWS::S3::Bucket",
      "replacement": "False",
      "moduleInfo": null,
      "details": [
        {
          "target": {
            "name": null,
            "requiresRecreation": "Never",
            "attribute": "Tags"
          },
          "causingEntity": null,
          "evaluation": "Static",
          "changeSource": "DirectModification"
        }
      ],
      "changeSetId": null,
      "scope": [
        "Tags"
      ]
    },
[...]

PR #600, opened 2021-06-18
feat: Define `SSMRunCommandPolicy` in the CDK stack

akash1810:
Builds on https://github.com/guardian/amigo/pull/598.

## What does this change?

Moving the `SSMRunCommandPolicy` resource into the CDK stack.

This is chosen because it's fairly easy to test if it works: if successful, we can continue to connect to an instance using ssm-scala.

AMIgo's stack defines two additional actions over the standard `GuSSMRunCommandPolicy`:
  1. ssm:StartSession
  2. ssm:TerminateSession

Slightly annoyingly, `GuSSMRunCommandPolicy` is a singleton and doesn't accept additional input. Therefore, we have to add these actions into a new statement on the policy.

`GuSSMRunCommandPolicy`'s `Roles` array is kinda hilarious (see snapshot test change). The good news is it's done by AWS CDK magic, so should work!

## How to test

- Perform a full deploy of this branch on CODE
- Attempt to connect to one of the instances (`ssm ssh -t amigo,CODE -p deployTools --newest -x`) and witness it succeed

We move closer to a CDK only template.

### DIFF
--- a/cdk/lib/amigo.ts
+++ b/cdk/lib/amigo.ts
@@ -1,8 +1,11 @@
 import path from "path";
+import type { CfnRole } from "@aws-cdk/aws-iam";
+import { Effect, PolicyStatement, Role } from "@aws-cdk/aws-iam";
 import { CfnInclude } from "@aws-cdk/cloudformation-include";
 import type { App } from "@aws-cdk/core";
 import type { GuStackProps, GuStageParameter } from "@guardian/cdk/lib/constructs/core";
 import { GuStack } from "@guardian/cdk/lib/constructs/core";
+import { GuSSMRunCommandPolicy } from "@guardian/cdk/lib/constructs/iam";
 
 const yamlTemplateFilePath = path.join(__dirname, "../../cloudformation.yaml");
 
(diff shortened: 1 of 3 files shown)


cdk/lib/amigo.ts
```diff
@@ -1,20 +1,39 @@
...
+    const rootRole = Role.fromRoleArn(this, "RootRole", cfnRootRole.attrArn);
+
+    const ssmPolicy = GuSSMRunCommandPolicy.getInstance(this);
+
+    // TODO Can GuSSMRunCommandPolicy expose its default `PolicyStatement` to easily add actions to it?
+    ssmPolicy.addStatements(
+      new PolicyStatement({
+        effect: Effect.ALLOW,
+        resources: ["*"],
+        actions: ["ssm:StartSession", "ssm:TerminateSession"],
```
jacobwinch, line 34: Any idea why AMIgo needs these extra permissions? If we're not sure, we could try dropping them and seeing if anything breaks in `CODE`?

akash1810: These have been purposefully added in https://github.com/guardian/amigo/pull/526 and https://github.com/guardian/amigo/pull/538. For this reason, I kind of like that they're expressed in new policy statements. An inline comment would be helpful! Will add.

jacobwinch: Thanks!

jacobwinch (approved):

PR #621, opened 2021-06-24
feat: Add `diff` script to easily compare against AWS

akash1810:
## What does this change?

The `cdk diff` command will compare a CDK stack against one running in AWS.

By default, `cdk` will use the `id` argument as the AWS stack name. That is, by default `cdk` will try to compare against the AWS stack "AMIgo".

This change creates a script that will read the stack name from an envvar instead, wrapping it all up in a new script for ease.

You can run `./script/diff` to compare against the CODE stack in AWS, or `./script/diff --prod` to compare against the PROD stack in AWS.

The script will also check your credentials as `cdk` seems to just hang if your credentials are invalid.

<details>
<summary>Example output</summary>

```console
➜ ./script/diff
yarn run v1.22.10
warning package.json: No license field
$ cdk diff --path-metadata false --version-reporting false
Stack AMIgo (amigo-CODE)
There were no differences
✨  Done in 3.62s.
```

</details>

## How to test

- Checkout this branch
- Run `./script/diff` and observe the output

Local development is that much easier.

### DIFF
--- a/cdk/script/diff
+++ b/cdk/script/diff
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+STACK_NAME=amigo-CODE
+PROFILE=deployTools
+
+for arg in "$@"; do
+  if [ "$arg" == "--prod" ]; then
+    STACK_NAME=amigo-PROD
+    shift
+  fi
+done
+
+checkCredentials() {
+  STATUS=$(aws sts get-caller-identity --profile ${PROFILE} 2>&1 || true)
+
+  if [[ ${STATUS} =~ (ExpiredToken) ]]; then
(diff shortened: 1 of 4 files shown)


cdk/script/diff
```diff
@@ -0,0 +1,32 @@
...
+PROFILE=deployTools
+
+for arg in "$@"; do
+  if [ "$arg" == "--prod" ]; then
+    STACK_NAME=amigo-PROD
+    shift
+  fi
+done
+
+checkCredentials() {
```
jacobwinch, line 15: Great, thanks for adding this (although it's a shame that we need to!)

jacobwinch (approved):
💯

PR #627, opened 2021-06-25
feat: Migrate policy for ImageCopier

akash1810:
Builds on #598.

Requires #626.

## What does this change?

I don't fully understand the implementation of the image copier feature! In this change I'm (somewhat blindly) moving the policy from the YAML template into the CDK stack. It should be a no-op.

## How to test

As mentioned above, I don't really understand this feature to conclusively say. Help wanted!

We can see from the snapshot diff, however, that the only change is where the policy gets defined. That is, I'm confident this is a no-op.

We move closer to a CDK only template.

### DIFF
--- a/cdk/lib/__snapshots__/amigo.test.ts.snap
+++ b/cdk/lib/__snapshots__/amigo.test.ts.snap
@@ -80,24 +80,6 @@ Object {
           "Statement": Array [
             Object {
               "Action": Array [
-                "s3:GetBucketPolicy",
-                "s3:PutBucketPolicy",
-              ],
-              "Effect": "Allow",
-              "Resource": Object {
-                "Fn::Join": Array [
-                  "",
-                  Array [
-                    "arn:aws:s3::*:",
-                    Object {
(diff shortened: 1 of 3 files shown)


jacobwinch (approved):

jacobwinch:
> As mentioned above, I don't really understand this feature to conclusively say. Help wanted!

I'm also a bit confused by this. I think that the PR which introduced this functionality (and these IAM permissions) is https://github.com/guardian/amigo/pull/164.

I think we [upload a single build artifact](https://github.com/guardian/amigo/blob/453e073b0e34db861f0a002eec5553cacb12017c/riff-raff.yaml#L21-L27) and then allow lambdas in different accounts to make use of this, but I'm not quite sure how attaching these permissions to AMIgo's app role helps us with that 😕

akash1810:
> Edit: I think it's allowing this code to work.

Ah, good spot. I think this implementation is still a little odd. Creating AWS resources in application code feels like an anti-pattern to me. I won't tug on that thread in this PR though 😅 .

PR #628, opened 2021-06-25
feat: Migrate Packer instance profile policy statement

akash1810:
Builds on #598.

Requires #627.

## What does this change?

Migrates the Packer SSM policy from the YAML template to the CDK stack.

## How to test

If this change isn't correct, I think the result would be an inability to bake any recipe. So, to test, try to bake a recipe?

We move closer to a CDK only template - the YAML defined `AmigoAppPolicy` is completely migrated 🎉 .

### DIFF
--- a/cdk/lib/amigo.ts
+++ b/cdk/lib/amigo.ts
@@ -6,7 +6,7 @@ import { CfnInclude } from "@aws-cdk/cloudformation-include";
 import type { App } from "@aws-cdk/core";
 import { Stage } from "@guardian/cdk/lib/constants";
 import type { GuStackProps, GuStageParameter } from "@guardian/cdk/lib/constructs/core";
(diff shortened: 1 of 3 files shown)


cdk/lib/amigo.ts
```diff
@@ -98,13 +99,27 @@ export class AmigoStack extends GuStack {
           actions: ["s3:GetBucketPolicy", "s3:PutBucketPolicy"],
           resources: [`arn:aws:s3::*:${GuDistributionBucketParameter.getInstance(this).valueAsString}`],
         }),
+
+        /*
+        See See https://github.com/guardian/amigo/pull/526
```
jacobwinch, line 104: Minor: ```suggestion
        See https://github.com/guardian/amigo/pull/526
```

jacobwinch (approved):

PR #631, opened 2021-07-01
feat: Migrate `PackerSecurityGroup` into CDK stack

akash1810:
## What does this change?

This change moves the YAML defined `PackerSecurityGroup` into the CDK stack. There is ultimately one change being made, the `AWS::EC2::SecurityGroup Egress` gets a description. The [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html#cfn-ec2-security-group-rule-description) state this isn't a replacement operation 🎉 .

## How to test

Deploy and bake - the bake should start.

Moving closer to a CDK only defined stack.


Nirvikalpa108 (approved):
Tested and working nicely 😀

PR #633, opened 2021-07-02
feat: Expose build info on `/healthcheck`

akash1810:
## What does this change?

The build info has useful information that aides debugging. Specifically the commit hash which helps one determine exactly what version of the app is deployed.

This doesn't impact the AWS healthcheck as the status code remains the same - only the response body changes.

## How to test

- Deploy this branch
- Visit `/healthcheck`
- Witness some lovely JSON

## Images

image

### DIFF
--- a/app/controllers/RootController.scala
+++ b/app/controllers/RootController.scala
@@ -1,13 +1,14 @@
 package controllers
 
+import amigo.BuildInfo
 import com.gu.googleauth.GoogleAuthConfig
-
+import play.api.libs.json.Json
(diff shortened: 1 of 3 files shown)


Nirvikalpa108 (approved):
Nice 👍🏽 
I guess it's not worth making the build time readable is it?

akash1810:
> I guess it's not worth making the build time readable is it?

Great idea! I've added a new commit for this. Build time will now be something like `"2021-07-02T08:43:23.909Z"`, much more useful!

PR #632, opened 2021-07-01
feat: Use the `GuPlayApp` pattern

akash1810:
## What does this change?

This change uses the `GuPlayApp` pattern to define the ASG, ALB etc for AMIgo 🎉 . It's pretty big, I'm struggling to break it down - help wanted!

With this change we are creating a new load balancer that will run in parallel to the current one. This will allow us to perform a zero-downtime migration via a three step dance:
  1. Run two load balancers
  2. Update the CNAME to point to the new load balancer
  3. Remove the old load balancer

There are no deletions being made in this change (the change set JSON is below), so we should not incur any downtime.

It's worth noting that instances launched in the new ASG are not configured exactly the same as the current instances. The current instances can be [reached directly on port 9000](https://github.com/guardian/amigo/blob/ad138c93b63b5fa01b8f44399ba1d3cc17a69654/cloudformation.yaml#L163-L166) as well as via the load balancer. The new instances only provide access to the new instances via the load balancer. Obviously, you can still connect to the instances via SSM.

<details>
<summary>Change set JSON</summary>

```json
[
  {
    "resourceChange": {
      "logicalResourceId": "AppPolicyF941AEC5",
      "action": "Modify",
      "physicalResourceId": "amigo-AppP-1KNFT10DMTIKV",
      "resourceType": "AWS::IAM::Policy",
      "replacement": "False",
      "moduleInfo": null,
      "details": [
        {
          "target": {
            "name": "Roles",
            "requiresRecreation": "Never",
            "attribute": "Properties"
          },
          "causingEntity": null,
          "evaluation": "Dynamic",
          "changeSource": "DirectModification"
        },
        {
          "target": {
            "name": "Roles",
            "requiresRecreation": "Never",
            "attribute": "Properties"
          },
          "causingEntity": "InstanceRoleAmigo75944747",
          "evaluation": "Static",
          "changeSource": "ResourceReference"
        }
      ],
      "changeSetId": null,
      "scope": [
        "Properties"
      ]
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
  {
    "resourceChange": {
      "logicalResourceId": "AutoScalingGroupAmigoASG5BE72894",
      "action": "Add",
      "physicalResourceId": null,
      "resourceType": "AWS::AutoScaling::AutoScalingGroup",
      "replacement": null,
      "moduleInfo": null,
      "details": [],
      "changeSetId": null,
      "scope": []
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
  {
[...]

PR #634, opened 2021-07-02
feat: Add the instanceId as a log marker

akash1810:
## What does this change?

This will allow us to determine which instance is sending logs, which is of particular use for the YAML -> CDK migration.

There are a lot of whitespace changes in this diff, so it'll be easier to review [w/out them](https://github.com/guardian/amigo/pull/634/files?diff=split&w=1).

## How to test

- Deploy this branch
- Check logs, they should have a new marker

This will allow us to determine if the new instances launched in #632 have the correct permissions.

## Images

image

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -2,32 +2,32 @@ package components
 
 import akka.stream.scaladsl.Source
 import akka.typed._
-import com.amazonaws.{ AmazonClientException, AmazonWebServiceRequest, ClientConfiguration }
-import com.amazonaws.auth.{ AWSCredentialsProviderChain, InstanceProfileCredentialsProvider }
+import com.amazonaws.{AmazonClientException, AmazonWebServiceRequest, ClientConfiguration}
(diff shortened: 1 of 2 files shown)


app/components/AppComponents.scala
```diff
@@ -66,10 +66,13 @@ class AppComponents(context: Context)
     with I18nComponents
     with Loggable {
 
+  val awsInstance = new AwsInstanceImpl(SysOutLogger)
+
   val identity = {
     import com.gu.cm.PlayImplicits._
-    Identity.whoAmI("amigo", context.environment.mode)
+    new InstanceDescriber("amigo", context.environment.mode, awsInstance, SysOutLogger).whoAmI
```
akash1810, line 73: This is a small efficiency change as we can reuse `awsInstance` above rather than creating a [new one](https://github.com/guardian/configuration-magic/blob/3b9c5f093ae92e682906b7cc1cd00255785b1a06/core/src/main/scala/com/gu/cm/Identity.scala#L61).

stephengeller (approved):
Looks good! @akash1810 has tested this change out in AWS and seen it succeed so happy to approve this, as the changes are expected.

PR #635, opened 2021-07-02
chore: Remove the YAML stack

akash1810:
Requires #632.

## What does this change?

In #632 we have migrated to the GuCDK pattern `GuPlayApp`.

This change removes the old infrastructure as it's not receiving any traffic. This completes the YAML to CDK migration 🎉 .

## How to test

Oof, this is an interesting question. I guess update the stack (after confirming the old load balancer isn't serving any traffic as the load balancer will be removed).

The app should continue to function.


stephengeller (approved):
Woohoo! Nice one.

akash1810:
#632 has been merged and DNS has been updated and propagated. This should be safe to merge now 🎉 . I'll update the stack manually though first.

PR #636, opened 2021-07-19
Tweaks to ssh reserve space

jorgeazevedo:
## What does this change?

This changes 3 things
1. moves image files to `/var/lib/reserved-disk`
2. Reserves 50mb instead of 10mb
3. Retains `/var/lib/amazon/ssm`'s original permissions, which are `drwxr-x---`

## How to test

- [x] Create an AMI with this role
- [x]  Create an instance of this AMI
- [x]  Use ssm ssh to get onto the instance
- [x]  Run `fallocate -l $(($(df . --output=avail | tail -1) * 1024)) filler` to fill the root partition
- [x]  Try to use ssm ssh again from another local terminal. It should work

### DIFF
--- a/roles/ssh-reserve-disk-space/tasks/main.yml
+++ b/roles/ssh-reserve-disk-space/tasks/main.yml
@@ -50,4 +50,4 @@
     path: /var/lib/amazon/ssm
     owner: root
     group: root
-    mode: '0700'
+    mode: '0750'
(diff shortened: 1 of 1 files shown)


davidfurey:
Thanks Jorge, I was just about to start making these changes in response to the feedback from @sihil.  Pleasantly surprised to see you've already done it!

jorgeazevedo:
Ok, I've finally tested this on an actual instance with a bake from amigo code and it works as expected!

davidfurey (approved):

PR #638, opened 2021-07-21
Replace use of bintray repository in teamcity agent role

jfsoul:
## What does this change?

The amigo bakes for the teamcity-agent recipe had started to fail, which from the logs we could easily spot was due to the teamcity-agent role utilising a (now sunsetted) bintray repo for downloading sbt. This PR updates that role using the instructions from https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Linux.html. Specifically, we switch the repo location and add an additional one based on the linked guide.  Couple of other small changes:

- The keyserver connection is now secure over the `hkps` protocol.
- I've moved the `sbt.yml` task into the `main.yml`, which seems to be the expectation of amigo currently (for better or worse). The effect of this is just that you can view the entire task configuration in the UI (supplemental tasks are omitted). It'd be possible instead to refactor the app to parse multiple task files and display them all in the UI, but it didn't seem worth it right now.

## How to test

I have run a new bake for the only recipe that utilises the updated role on Amigo CODE. I then used the generated AMI in the test cloud profile on teamcity in order to run a build of Amigo itself.  Everything worked as expected.

Bakes for teamcity-agent work again, which allows us to more easily manually update the AMI used by teamcity (and soon hopefully automate it)

It's possible something unexpected could happen when building certain projects using agents with the new role, but I think the risk should be minimal and could roll back fairly easily by switching out the AMI used by teamcity to an older one.

## Images

N/A

## Accessibility

N/A


jacobwinch (approved):
Great work 💯

PR #640, opened 2021-08-10
Add role that installs Python3.7 and virtual environment

michel-ds:
## What does this change?
Adds an ansible role that installs:
- `python3.7`
- `pip3`
- `virtualenv` for Python 3

Installs Python 3.7 as a longer support option compared to the system Python 3.6 (EOL: [27 Jun 2023 vs. 23 Dec 2021](https://endoflife.date/python)).
I am open to replace it with Python 3.8 as an even longer support option.

## Testing
- [x] Tested ansible role as set out in `README.md` (including manual check that all 3 installed packages run on the provisioned Vagrant box)

This role does **not** overwrite the Python 3.6 version installed via `apt-get install python3`, hence it will not conflict with any of the other Python 3 roles in this repo (`pip3`, `pip3-packages`).
The following symlink will remain intact:
`/usr/bin/python3` --> `/usr/bin/python3.6`

### DIFF
--- a/roles/python3.7-pip-virtualenv/tasks/main.yml
+++ b/roles/python3.7-pip-virtualenv/tasks/main.yml
@@ -1,16 +1,6 @@
 ---
-- name: install pip
+- name: install pip, python 3.7 and virtualenv
   apt:
-    name: ['python3-pip']
-    update_cache: yes
-    state: latest
-- name: install python3.7
-  apt:
-    name: ['python3.7']
-    update_cache: yes
-    state: latest
-- name: install virtualenv
-  apt:
-    name: ['virtualenv']
+    name: ['python3-pip', 'python3.7', 'virtualenv']
     update_cache: yes
     state: latest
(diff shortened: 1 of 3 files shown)


jorgeazevedo (approved):

roles/python3.7-pip-virtualenv/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: install pip
+  apt:
+    name: ['python3-pip']
```
jorgeazevedo, line 4: I wonder if this could be an array like `['python3-pip', 'python3.7', 'virtualenv']` ? I saw an example here https://github.com/guardian/amigo/blob/3843f69bcc212ff725084bbb359347c5dadab6fe/roles/wazuh-agent/tasks/main.yml#L5

michel-ds: @jorgeazevedo Yep, good point. I have do so now and tests ran successfully. https://github.com/guardian/amigo/pull/640/commits/29c3712f2522fa35bf81137f67536dfb15d16476

jorgeazevedo (approved):

PR #641, opened 2021-08-11
[role: `browser-testing`] make version of Chrome and chromedriver configurable (falling back to 'latest')

twrichards:
Currently [guardian/editorial-tools-production-monitoring](https://github.com/guardian/editorial-tools-production-monitoring) breaks with Chrome 92 (see https://trello.com/c/8tvMxMRX/2446-investigate-why-prodmon-breaks-w-the-latest-ami-and-fix for all the things we've tried in order to resolve). As such we had to remove the `ami-cloudformation-parameter` deployment step (see https://github.com/guardian/editorial-tools-production-monitoring/pull/281) so prodmon is not using the latest AMI, as it includes Chrome 92 (because it is 'latest').

## What does this change?
This PR makes the version of Chrome (and chromedriver) configurable (falling back to _latest_ if unspecified), so that we can specify Chrome 91 (which is the company wide version at the moment anyway) until there's a compatible & stable version of selenium available to work work with Chrome 92 (or above).

## How to test
Already tested...

- https://amigo.code.dev-gutools.co.uk/recipes/browser-testing-bionic/bakes/1 (without config, i.e. latest)
image

- https://amigo.code.dev-gutools.co.uk/recipes/browser-testing-bionic/bakes/2 (with config, downloads chrome 91)
image
image

We can revert https://github.com/guardian/editorial-tools-production-monitoring/pull/281 so at least we have up to date(ish) AMIs (i.e. latest OS etc. security patches) even if we don't have latest Chrome.

### DIFF
--- a/roles/browser-testing/README.md
+++ b/roles/browser-testing/README.md
@@ -4,6 +4,8 @@ This feature installs Google Chrome, Chromedriver and the virtual frame buffer, 
 
 The install script does the following:
  - Installs Xvfb
- - Installs the latest version of Google Chrome and associated packages
- - Installs latest Chromedriver version
+ - Installs the version of Google Chrome specified by `chrome_version` (installs latest if blank) and associated packages
+   <br/>_see https://www.ubuntuupdates.org/package/google_chrome/stable/main/base/google-chrome-stable for available options_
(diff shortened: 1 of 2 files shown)


sihil (approved):
LGTM

PR #642, opened 2021-08-11
Revert PR #640

michel-ds:
## What does this change?
Remove bespoke apt package installation role (`python3.7-pip-virtualenv`) as the same can be achieved with the [`packages` role](https://amigo.gutools.co.uk/roles#packages) as per conversation: https://chat.google.com/room/AAAAag0I08g/m7WVe37Wqls


jfsoul (approved):
Thanks! 👍

PR #659, opened 2021-12-03
Upgrade wazuh-agent to 4.x

louishather:
## What does this change?

Wazuh agents currently use version 3.13.1, and while they shouldbe upgraded as a matter of course, version 4.x offers a number of benefits, specifically around the lifecycle of agents and managers.

To change the version the apt source for wazuh-agent was set to 4.x, rather than 3.x. 

The role pulls SSM parameters to retrieve agent config, and specific wazuh-agent version. If the agent version parameter isn't changed when this change is merged, AMIs that use the wazuh-agent role will fail to build.

Applying this change needs coordination to change the SSM parameter values simultaneously.

### DIFF
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -17,7 +17,7 @@
 
 - name: Add wazuh apt repository
   apt_repository: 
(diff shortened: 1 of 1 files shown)


adamnfish:
Duplicated of https://github.com/guardian/amigo/pull/660

PR #663, opened 2021-12-08
Upgrade wazuh-agent to 4.x returns

louishather:
## What does this change?

This is another attempt at upgrading the wazuh-agent version, as #662 reverted the previous attempt. This was reverted as the wazuh-agent didn't start up, so there must have been changes to its default systemd configuration. There's now a task to explicitly enable the agent was part of the wazuh-agent role.  

Rather than use the script to retrieve the (non-secret) version from SSM, I've included it in the roles defaults file. This will mean future version upgrades can be tested without changes to the role or parameters in SSM - we can instead change the value for the base images. 

From #659 

> Wazuh agents currently use version 3.13.1, and while they shouldbe upgraded as a matter of course, version 4.x offers a number of benefits, specifically around the lifecycle of agents and managers.
> 
> To change the version the apt source for wazuh-agent was set to 4.x, rather than 3.x.
> 
> The role pulls SSM parameters to retrieve agent config, and specific wazuh-agent version. If the agent version parameter isn't changed when this change is merged, AMIs that use the wazuh-agent role will fail to build.

## How to test

Deploy an image built with Amigo CODE (after deploying this branch) and check the wazuh-agent is performing correctly on version 4.x. 

This should not be merged until this branch has been tested on amigo CODE, as it may cause new instances to not report to wazuh for the duration of their existence.

### DIFF
--- a/roles/wazuh-agent/defaults/main.yml
+++ b/roles/wazuh-agent/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 stage: PROD
+agent_version: 4.1.5-1
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -129,4 +129,4 @@
 - name: Enable wazuh-agent service
   service:
     name: wazuh-agent
-    state: enabled
+    enabled: yes
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -43,14 +43,9 @@
   register: manager_address
   failed_when: "'error' in manager_address.stderr"
 
-- name: Get agent version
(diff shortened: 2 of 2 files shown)


jorgeazevedo:
Apologies this is taking a while to review! I'm afraid of merging this until I understand why prism instances failed to launch the last time around. Did you see any behaviour like that in your tests?

jorgeazevedo:
And... by sheer coincidence, looks like many of our bakes are broken due to an issue related to python. See https://chat.google.com/room/AAAAag0I08g/lN-uKQex5RA and https://github.com/guardian/amigo/pull/665

I was just now going to try to reproduce the prism issue, but I can't bake anything, so I think we're blocked until this python issue is sorted : -\

louishather:
I've been able to test an AMI built with this branch - the AMIs are built correctly and the wazuh agent registers to our cluster.

Good to go from my end, whenever someone's able to review this

jfsoul (approved):

PR #670, opened 2021-12-14
First draft of an AMI-level mitigation for CVE-2021-44228

adamnfish:
## What does this change?

Add a role to set an environment variable that will (partially) disable the Log4J2 feature at the root of CVE-2021-44228.

We are adding this to a new role that we intend to add to all our base images. This ensures that this feature is globally enabled, to catch our Java AMIs as well as anything else that uses Java surprisingly.

## How to test

We'll have to bake an image and check that this environment variable is correctly configured on a box that is started up with the newly-built AMI. We'll then also check that an application started the normal way also sees the environment variable.

We'll need to make sure that this works before we can consider it a true mitigation but this is only a defence-in-depth approach to tackling the Log4J issue, so the risk here is small.


jfsoul (approved):

roles/security-patches/tasks/main.yaml
```diff
@@ -0,0 +1,5 @@
+## Automatically disable Log4J2 vulnerability CVE-2021-44228
+- name: Disable Log4J vulnerable features
+  copy:
+    src: 99-log4j.conf
```
jfsoul, line 4: did you intentionally have 99- prefix for the src and 10- prefix for the dest?

jfsoul:
Given that CVE-2021-45046 has been upgraded to a critical vulnerability, I think this approach has lower value and we're prioritising #673 over it for now. We may come back to this if still deemed useful, or the other approach fails.

PR #671, opened 2021-12-15
Update agent to use log4j v2.16.0

jorgeazevedo:
## What does this change?

Updates the kinesis agent to the latest version, which was released specifically to bump the log4j version. See https://github.com/awslabs/amazon-kinesis-agent/releases

## How to test

I'm not sure how to test this actually

This role is used by dotcom rendering, most of MAPI's infrastructure and a few others. If this is broken, it might break bakes, or break the apps ability to ship logs to kinesis.

### DIFF
--- a/roles/aws-kinesis-agent/tasks/main.yml
+++ b/roles/aws-kinesis-agent/tasks/main.yml
@@ -9,15 +9,15 @@
   file: path=/opt/aws-kinesis-agent/ state=directory mode=0755
 
 - name: Download Amazon Kinesis Agent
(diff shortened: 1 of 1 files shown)


jacobwinch (approved):
> I'm not sure how to test this actually

We could:

1. Clone one of MAPI's recipes
2. Bake to pick up these role changes
3. Deploy the service to `CODE` and check that logs still show up in Central ELK

But as this is a patch update I think we should be reasonably confident that it will continue working 🤞

jorgeazevedo:
Thanks @jacobwinch , I'll do that!

PR #673, opened 2021-12-16
Add security-patch role to mitigate CVE-2021-44228 and CVE-2021-45046

jfsoul:
## What does this change?
Add a role to set an environment variable that adds a [java agent](https://github.com/corretto/hotpatch-for-apache-log4j2) to all java processes which then disables the Log4J2 feature at the root of CVE-2021-44228 and CVE-2021-45046

We are adding this to a new role that we intend to add to all our base images. This ensures that this feature is globally enabled, to catch our Java AMIs as well as anything else that uses Java surprisingly.

image

## How to test
We'll have to bake an image and check that this environment variable is correctly configured on a box that is started up with the newly-built AMI. We'll then also check that an application started the normal way also sees the environment variable.

There could be risks associated with adding a java agent to all running processes and as such we'll do some testing on different kinds of apps (elasticsearch, scala web app) first. However there is some comfort in AWS enabling the same agent on all lambda java runtimes. There's also a risk this environment variable could be overwritten, which is unavoidable AFAICT.


jfsoul:
Tested successfully on Amigo/Prism CODE via systemd (i.e. the usual way) and via running the prism `java` command directly from the command line.

jfsoul:
Also tested on Ubuntu 20.04

roles/security-patches/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  copy:
+      src: log4j-systemd.conf
+      dest: /etc/systemd/system.conf.d/99-log4j-hotpatch.conf
+
+- name: Set log4j hotpatch environment outside systemd
+  ansible.builtin.lineinfile:
+      path: /etc/environment
+      create: yes
+      insertafter: EOF
+      line: JAVA_TOOL_OPTIONS="-javaagent:/usr/local/log4j-hotpatch/Log4jHotPatch.jar"
```
adamnfish, line 22: Should this do an "append" rather than a "set", on the `JAVA_TOOL_OPTIONS` environment variable?

roles/security-patches/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+- name: Create systemd conf directory for log4j hotpatch environment
+  file: path=/etc/systemd/system.conf.d state=directory
+
+- name: Set systemd environment for log4j hotpatch
+  copy:
+      src: log4j-systemd.conf
+      dest: /etc/systemd/system.conf.d/99-log4j-hotpatch.conf
+
+- name: Set log4j hotpatch environment outside systemd
+  ansible.builtin.lineinfile:
```
adamnfish, line 18: Does this work without the `ansible.builtin` prefix (based on how copy and file seem to)?

adamnfish (approved):
Couple of little things to check, but we've chatted IRL about the append/set issue so I reckon this is fine.

roles/security-patches/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+## Automatically disable Log4J2 vulnerability CVE-2021-44228 and CVE-2021-45046
+
+- name: Create /usr/local/log4j-hotpatch directory
+  file: path=/usr/local/log4j-hotpatch state=directory
```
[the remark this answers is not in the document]
jfsoul: Yeah, we can do that, only this way due to copy+paste really

roles/security-patches/tasks/main.yml
jfsoul: I believe so, as we're not doing anything complicated like loading custom modules. I'll have to try it

roles/security-patches/tasks/main.yml
jfsoul: As discussed elsewhere, I haven't been able to get the interpolation to work, and we deem the risk to be pretty low given a github org search for the variable name return local development scripts only.

PR #674, opened 2021-12-17
Fix base image End of Life label

MarSavar:
While pairing with @philmcmahon earlier today we spotted that the label for end of life images was not displaying correctly.

This is because the `label-error` CSS class does not exist. This PR addresses the problem and makes the label readable.

## Before
image

## After
image


jorgeazevedo (approved):
Great spot!

PR #678, opened 2021-12-22
Add JDK 17 as a role

mchv:
## What does this change?

JRE17 has been supported by Scala for [some time now](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#version-compatibility-table) and notably the only version which works natively on [M1 ARM 64 architecture](http://openjdk.java.net/jeps/391).

This adds a role that should be a simple swap out for recipes switching from version `8` or `11`.

## How to test
This impacts baked AMIs so best tested by converting an AMI over.

See #548

### DIFF
--- a/roles/java17/spec/java17_spec.rb
+++ b/roles/java17/spec/java17_spec.rb
@@ -1,0 +1,4 @@
+
+describe command('java -version') do
+  its(:stdout) { should match /java version "17\..*/ }
+end
--- a/roles/java17/meta/main.yml
+++ b/roles/java17/meta/main.yml
@@ -1,0 +1,4 @@
+---
+dependencies:
+  - role: apt
+    when: ansible_os_family == "Debian"
--- a/roles/java17/tasks/main.yml
+++ b/roles/java17/tasks/main.yml
@@ -1,0 +1,18 @@
+---
+- name: Install Java 17 JRE and JDK
+  apt:
+    name:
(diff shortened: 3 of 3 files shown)


mchv:
Please note that `Play` does not support yet JDK 17

roles/java17/spec/java17_spec.rb
```diff
@@ -0,0 +1,4 @@
+
+describe command('java -version') do
+  its(:stdout) { should match /java version "17\..*/ }
+end
```
jfsoul, line 4: Interestingly @adamnfish and I noticed that these specs are not run anywhere yet. They were added in expectation that we'd get round to supporting them one day.

mchv: Thanks, that is useful to know! I have now found #6 and I think it will be a great thing to look at.

nicl (approved):
Have tested this on Focus/ARM on CODE and the bake succeeds. I haven't tested the resultant AMI but am relatively confident this is fine as it is a straightforward port of the Java 11 role.

Thanks @mchv !

philmcmahon:
@nicl around the time of the ARM stuff some people found that AWS's JDK https://aws.amazon.com/about-aws/whats-new/2021/09/amazon-corretto-17-now-available/ solved some of the performance problems seen on standard JDK11 - it might be worth doing a role for that as well, or instead of this one

PR #680, opened 2022-01-06
Add corretto 17 role

nicl:
## What does this change?

Adds Corretto's Java 17. See also https://github.com/guardian/amigo/pull/678.

We should probably go further than this and actually express a clear preference over which Java distro to use. But that is beyond the scope of this PR.

cc @philmcmahon .

## How to test

Tested locally.

Screenshot 2022-01-06 at 16 49 10

### DIFF
--- a/roles/java17corretto/tasks/main.yml
+++ b/roles/java17corretto/tasks/main.yml
@@ -1,0 +1,28 @@
+---
+- name: Add AWS Corretto signing key
+  apt_key:
+    url: "https://apt.corretto.aws/corretto.key"
+    state: present
+
+- name: Add AWS Corretto repo
+  apt_repository:
+    repo: 'deb https://apt.corretto.aws stable main'
+    state: present
+
+- name: Install Java 17 Corretto JDK
+  apt:
+    name:
+    - java-17-amazon-corretto-jdk
+    state: latest
+  when: ansible_os_family == "Debian"
+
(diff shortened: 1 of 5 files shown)


roles/java17corretto/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  apt_repository:
+    repo: 'deb https://apt.corretto.aws stable main'
+    state: present
+
+- name: Install Java 17 Corretto JDK
+  apt:
+    name:
+    - java-17-amazon-corretto-jdk
+    state: latest
+  when: ansible_os_family == "Debian"
```
jfsoul, line 17: does this condition want adding to all statements? I don't recall off the top of my head but probably there's a way in ansible to include a set of tasks/file with one condition statement

jfsoul (approved):
Not 100% sure what this task would do on non-debian at the moment without a small tweak, but otherwise looks fine!

roles/java17corretto/tasks/main.yml
philmcmahon: I think this line doesn't need to be there as we're not offering a redhat alternative. I don't think we're using amazon linux for anything at the moment so seems reasonable to only support debian in this role

nicl:
@jfsoul yep. It's currently constrained to Debian only so shouldn't break anything at least.

PR #681, opened 2022-01-17
Upgrade to Play 2.6

jacobwinch:
## What does this change?

This PR upgrades to Play Framework to version 2.6, following the [migration guide](https://www.playframework.com/documentation/2.8.x/Migration26).

There are a number of significant changes between Play 2.5 and Play 2.6, so I've attempted to make minimal changes to other code/dependencies in order to keep this PR as small and easy to review as possible. That is, I've only touched other dependencies to workaround compiler/runtime errors. This PR is the first part of an effort to bring all of this project's dependencies up to date.

## How to test

I've deployed this to `CODE` and tested the following behaviour:

* Authentication
* Create base image
* Edit base image
* Delete base image
* Create recipe
* Clone recipe
* Edit recipe
* Delete recipe
* Bake an AMI
* Delete bake / AMIs

This is the first step towards bringing all dependencies for this project up to date. After this PR some dependencies will be a bit less out of date and other upgrades (e.g. [Scala version](https://www.playframework.com/documentation/2.8.x/Highlights26#Scala-2.12-support)) will be unblocked.

Yes, this is a significant upgrade and although I've done quite a bit of testing and fixed a number of bugs, it's possible that some things have been missed.

Part of the strategy behind moving to Play 2.6 without upgrading all other possible dependencies (e.g. Scala version) is to break the upgrades down into stages. Hopefully this will make it easier to spot/fix bugs if we introduce them. 

## To do 

- [ ] Add `play.http.secret.key` to `PROD` config **before merging**
- [ ] Delete `play.crypto.secret` from `CODE`/`PROD` config **after merging**


app/AppLoader.scala
```diff
@@ -5,20 +5,25 @@ import play.api.ApplicationLoader.Context
...
+      val components = new AppComponents(context)
+      log.info("Starting the scheduler")
+      components.scheduler.start()
+      components.applicationLifecycle.addStopHook { () =>
+        log.info("Shutting down scheduler")
+        Future.successful(components.scheduler.shutdown())
+      }
+      components.application
+    } catch {
+      case err: Throwable =>
```
jacobwinch, line 23: This ensures that errors which occur when the app first starts make it into the logs

app/components/AppComponents.scala
```diff
@@ -64,27 +67,43 @@ class AppComponents(context: Context)
...
     with AhcWSComponents
     with I18nComponents
-    with Loggable {
+    with Loggable
+    with AssetsComponents
+    with HttpFiltersComponents {
 
   val awsInstance = new AwsInstanceImpl(SysOutLogger)
 
+  val configurationMagicMode = context.environment.mode match {
```
jacobwinch, line 76: We use an old version of https://github.com/guardian/configuration-magic for config which is incompatible with Play 2.6 (I believe this is due to [Mode API changes](https://www.playframework.com/documentation/2.6.x/Migration26#Scala-Mode-changes)). As far as I can tell we cannot upgrade this library without also bumping Scala version (which I want to avoid in this first PR).

As `configuration-magic` is unsupported anyway, this PR just adapts a couple of functions from the library to work with Play 2.6 and brings them inline. 

I want to remove this library and use https://github.com/guardian/simple-configuration instead, as per our [recommendation](https://github.com/guardian/recommendations/blob/main/config.md). However, that involves moving all config from DynamoDB to Parameter Store, so I will make this a separate PR.

app/components/AppComponents.scala
```diff
@@ -94,7 +113,7 @@ class AppComponents(context: Context)
     ))
 
   // initialise logging
-  val elkLoggingStream = configuration.getString("elk.loggingStream")
+  val elkLoggingStream = configuration.get[Option[String]]("elk.loggingStream")
```
jacobwinch, line 116: [Config API changes in Play 2.6](https://www.playframework.com/documentation/2.6.x/Migration26#Scala-Configuration-API) mean that defaults are now often provided via https://github.com/guardian/amigo/blob/jw-play/conf/application.conf.

app/controllers/ControllerHelpers.scala
```diff
@@ -1,10 +1,6 @@
 package controllers
 
 import models.{ RoleId, CustomisedRole }
-import cats.syntax.either._
```
jacobwinch, line 4: `cats` was not explicitly declared in `build.sbt` and we started hitting errors when `scanamo` was upgraded. I've rewritten the code without `cats` rather than trying to find compatible version. In a lot of places we were just using it to get [right-biased Either](https://typelevel.org/cats/datatypes/either.html#syntax) anyway and we will get that for free when we upgrade Scala

app/event/Behaviours.scala
```diff
@@ -16,26 +16,22 @@ object Behaviours extends Loggable {
...
   /**
    * Broadcasts all incoming events to all event listeners
-   * and handles child failures by restarting the child
    */
   def broadcastEvents(eventListeners: Iterable[ActorRef[BakeEvent]]): Behavior[BakeEvent] = Full {
     case Msg(_, bakeEvent) =>
       for (listener <- eventListeners)
         listener ! bakeEvent
       Same
-    case Sig(_, f @ Failed(cause, child)) =>
```
jacobwinch, line 36: This restart behaviour is not present in later versions of the library

app/views/confirmBakeDelete.scala.html
```diff
@@ -9,7 +10,10 @@ <h1>Really delete @bakeId.recipeId bake @bakeId.buildNumber? </h1>
             <div class="panel-heading">Confirm delete</div>
             <div class="panel-body">
                 <p>Note that this will schedule the deletion of all encrypted copies of this bake within the next @bakeDeletionFrequencyMinutes minutes.</p>
-                <p><a data-href="@routes.BakeController.deleteBake(bakeId.recipeId, bakeId.buildNumber)" class="post btn btn-danger">Delete</a></p>
+                @helper.form(routes.BakeController.deleteBake(bakeId.recipeId, bakeId.buildNumber)) {
+                    @CSRF.formField
```
jacobwinch, line 14: There are a number of changes like this - see https://www.playframework.com/documentation/2.6.x/Migration26#CSRFFilter for more details

app/views/editBaseImage.scala.html
```diff
@@ -41,14 +43,13 @@
                 </label>
               </div>
               <div>
-                <input type="text" class="form-control" style="visibility:hidden" id="role-@role-variables" name="role-@role-variables" placeholder="Custom variables (optional) e.g. foo: bar, baz: wow">
+                <input type="text" class="form-control hidden" id="role-@role-variables" name="role-@role-variables" placeholder="Custom variables (optional) e.g. foo: bar, baz: wow">
```
jacobwinch, line 46: Inline styles are blocked by Play 2.6 - see https://www.playframework.com/documentation/2.6.x/Migration26#SecurityHeadersFilter

build.sbt
```diff
@@ -69,14 +68,16 @@ libraryDependencies ++= Seq(
...
   "com.beachape" %% "enumeratum" % "1.3.7",
-  "com.typesafe.akka" %% "akka-typed-experimental" % "2.4.2",
+  "com.typesafe.akka" %% "akka-typed-experimental" % "2.4.11",
   "com.typesafe.akka" %% "akka-agent" % "2.4.2",
   "com.gu" %% "configuration-magic-play2-4" % "1.3.0",
-  "com.gu" %% "play-googleauth" % "0.4.0",
-  "com.adrianhurt" %% "play-bootstrap3" % "0.4.5-P24",
+  "com.typesafe.play" %% "play-iteratees" % "2.6.1",
+  "com.typesafe.play" %% "play-iteratees-reactive-streams" % "2.6.1",
+  "com.gu" %% "play-googleauth" % "0.7.6",
```
jacobwinch, line 79: In later versions [it's necessary to use `play-secret-rotation` to avoid deprecation warnings](https://github.com/guardian/play-googleauth#versions). Again, it feels like it's worth tackling that in a separate PR.

app/views/showRecipe.scala.html
```diff
@@ -22,13 +23,24 @@ <h1>@recipe.id.value</h1>
...
+      <div class="btn-toolbar">
+        @helper.form(routes.BakeController.startBaking(recipe.id)) {
+          @CSRF.formField
+          <button class="post btn btn-primary">Bake!</button>
+        }
+        <a href="@routes.RecipeController.editRecipe(recipe.id)" class="btn btn-default">Edit</a>
+        @if(debugAvailable){
+          @helper.form(routes.BakeController.startBaking(recipe.id, debug = true)) {
+            @CSRF.formField
+            <button class="post btn btn-warning mr-1">Bake with debug enabled</button>
```
akash1810, line 35: What does the `mr-1` class do?

A button is more semantic than a link here 👍🏽 .

app/views/showRecipe.scala.html
```diff
@@ -22,13 +23,24 @@ <h1>@recipe.id.value</h1>
...
-      @if(debugAvailable){<a data-href="@routes.BakeController.startBaking(recipe.id, debug = true)" class="post btn btn-warning">Bake with debug enabled</a>}
-      <a href="@routes.RecipeController.deleteConfirm(recipe.id)" class="btn btn-danger">Delete...</a>
+      <div class="btn-toolbar">
+        @helper.form(routes.BakeController.startBaking(recipe.id)) {
+          @CSRF.formField
+          <button class="post btn btn-primary">Bake!</button>
+        }
+        <a href="@routes.RecipeController.editRecipe(recipe.id)" class="btn btn-default">Edit</a>
+        @if(debugAvailable){
+          @helper.form(routes.BakeController.startBaking(recipe.id, debug = true)) {
```
akash1810, line 33: It looks like other forms on this page are defined via `b3.form`, should this match?

conf/application.conf
```diff
@@ -1,10 +1,10 @@
-play.crypto.secret = "please overwrite me in prod"
+play.http.secret.key = "please overwrite me in prod"
 play.i18n.langs = [ "en" ]
 play.application.loader = AppLoader
 
 play.application.name = "amigo"
 
-aws.region = "eu-west-1"
+play.filters.headers.contentSecurityPolicy = "default-src 'self' https://maxcdn.bootstrapcdn.com/bootstrap/ https://code.jquery.com/ https://cdnjs.cloudflare.com/ajax/libs/showdown/1.4.3/ 'sha256-n5iMc+TYRqvKiJKPzco5gfPgoRaBlwBZqb+yNH4JTLo='"
```
akash1810, line 7: What does the sha256 represent?

conf/application.conf
```diff
@@ -14,16 +14,27 @@ google {
...
 
 packer {
-  ## You may need to set this in ~/.configuration-magic/amigo.conf when running locally
-  #vpcId = "vpc-foo"
-  #subnetId = "subnet-bar"
+  vpcId = null
+  subnetId = null
+  instanceProfile = null
+  securityGroupId = null
+  maxInstances = 5
```
akash1810, line 21: This looks like a new property? Is it worth documenting?

app/views/showRecipe.scala.html
jacobwinch: Yep, good idea!

app/views/showRecipe.scala.html
jacobwinch: I don't think it's actually doing anything useful. I was experimenting whilst trying to remove an annoying spacing issue that is introduced by this PR:

image

In the end I gave up because this button is only visible when running locally/in `CODE` anyway - I have removed the mysterious `mr-1` class now as I don't think it was helping here.

conf/application.conf
akash1810: So is this a sha256 hash of clipboard.min.js? CSP is fun...!

Not related to this PR - committing external JS isn't good practice, we should add a client-side build to this repo. This'll also mean we no longer pull in jquery from an external CDN.

conf/application.conf
akash1810: Ah! Thanks for that.

app/event/Behaviours.scala
jacobwinch: https://github.com/guardian/amigo/pull/683

app/components/AppComponents.scala
jacobwinch: https://github.com/guardian/amigo/pull/684

PR #682, opened 2022-01-18
Revert "Upgrade to Play 2.6"

jacobwinch:
Unfortunately we've noticed that some bakes are being marked as `TimedOut` even though the AMI is being created successfully behind the scenes. This is causing confusion and sending failure notifications to teams, so I'd like to revert this change whilst we investigate.


nicl (approved):
(👍 )

PR #683, opened 2022-01-20
Upgrade to Play 2.6 (attempt 2!)

jacobwinch:
## What does this change?

This is a second attempt at https://github.com/guardian/amigo/pull/681, which was reverted by https://github.com/guardian/amigo/pull/682.

There are a couple of important changes:

1. https://github.com/guardian/amigo/commit/a57d1fdcee36a5395ee00566612368880cf8cd0a updates the version of Akka and https://github.com/guardian/amigo/pull/683/commits/b2a48e1cf53db3dafe7297b7143e1381a64b0df1 reinstates some restart behaviour that was dropped [here](https://github.com/guardian/amigo/pull/681/files#r785926275). This is an attempt to ensure that updates to bake status and bake logs (in Dynamo) are always handled - we observed that both of these things stopped updating after  https://github.com/guardian/amigo/pull/681 had been deployed for a few hours, even though bakes were working correctly behind the scenes. 
2. https://github.com/guardian/amigo/commit/6171d827cf13e649b7e940c5c9957d5b58929701 - fixes an issue with real-time bake logs that I noticed whilst testing the previous commit.

## How to test

I have manually tested some bakes in `CODE` - it's hard to be sure whether this fixes the original issue properly without having another go at deploying this to `PROD`.

After the last attempt at deploying this change, there was a big spike in bakes (incorrectly) being [marked as timed out](https://logs.gutools.co.uk/s/devx/goto/7b518c10-7a10-11ec-99d2-79d743a54bd2). We will be able to monitor these logs and see if the problem happens again.

Yes, I'm not completely sure whether this will fix the issue, so it's possible that we will see a reoccurrence of the problem described above. The impact isn't too severe, so I think monitoring this change in `PROD` is OK?


app/event/Behaviours.scala
```diff
@@ -16,51 +16,61 @@ object Behaviours extends Loggable {
...
-  }
+  def broadcastEvents(eventListeners: Iterable[ActorRef[BakeEvent]]): Behavior[BakeEvent] = Behaviors
+    .supervise {
+      Behaviors.receiveMessage[BakeEvent] { bakeEvent =>
+        for (listener <- eventListeners)
+          listener ! bakeEvent
+        Behaviors.same
+      }
+    }
+    .onFailure(SupervisorStrategy.restart)
```
jacobwinch, line 37: I was a bit unsure where the supervision/restart strategy should be applied in this file - I think [these docs](https://doc.akka.io/docs/akka/2.5/typed/fault-tolerance.html#wrapping-behaviors) suggest that doing it here will be OK? Any feedback on this is welcome!

jacobwinch:
> Are these hashes likely to change over time? If so, it might be worth having an inline comment with how to calculate it.

I don't think people are editing this code [very frequently](https://github.com/guardian/amigo/blame/84791b6fb84b21bbbc4183fa6168941af3d1495f/app/views/showBake.scala.html#L117-L130!), but I've added a [comment](https://github.com/guardian/amigo/pull/683/commits/98180b6aa6fa2c34a16b8995c441dd2a528ec05f) just in case!

app/event/Behaviours.scala
jacobwinch: I tried this out with a simplified example and it didn't work as desired, hence https://github.com/guardian/amigo/commit/b2a48e1cf53db3dafe7297b7143e1381a64b0df1.

nicl (approved):
Have had the Akka-related changes kindly explained to me IRL by @jacobwinch and makes sense.

jacobwinch:
> After the last attempt at deploying this change, there was a big spike in bakes (incorrectly) being [marked as timed out](https://logs.gutools.co.uk/s/devx/goto/7b518c10-7a10-11ec-99d2-79d743a54bd2). We will be able to monitor these logs and see if the problem happens again.

This was deployed overnight and the problem didn't occur. I also checked a number of scheduled bakes manually and these all seem to have worked as expected.

PR #684, opened 2022-01-25
Replace configuration-magic with simple-configuration

jacobwinch:
## What does this change?

This PR removes `configuration-magic` and uses `simple-configuration` instead, as the former is no longer supported. This means that we are now reading private configuration from parameter store instead of Dynamo DB, and allows us to remove some IAM permissions. It also brings us in line with the [departmental recommendation](https://github.com/guardian/recommendations/blob/main/config.md) for config.

## How to test

I've deployed this to `CODE` and confirmed that I can still bake an AMI.

We have removed an out of date dependency, enabling easier Scala/Play upgrades in the future.

All of the configuration had to be ported from DynamoDB, so it's possible that there is an error with the `PROD` config that is not present in `CODE`. We'll only know for sure that this works by deploying it to `PROD`. I will leave the old config in DynamoDB for a while before removing it to allow us to rollback if necessary.


jacobwinch:
> Snyk is suggesting the version of simple-configuration being used is introducing transitive vulnerabilities. Do we want to release a new version of simple-configuration to resolve those?

Ah yes, thanks for pointing this out. We are using [the last version that was published for Scala 2.11](https://mvnrepository.com/artifact/com.gu/simple-configuration-ssm). Perhaps this is fixed in later versions that we can use once we've upgraded to 2.12 🤞 

I guess we can workaround it for now by pinning the version of the vulnerable transitive dependency...

app/AppLoader.scala
```diff
@@ -1,17 +1,23 @@
...
 
 import scala.concurrent.Future
-import scala.util.{ Failure, Success, Try }
 
 class AppLoader extends ApplicationLoader with Loggable {
   override def load(context: Context): Application = {
     new LogbackLoggerConfigurator().configure(context.environment)
     try {
-      val components = new AppComponents(context)
+      val identity = AppIdentity.whoAmI(defaultAppName = "myApp")
```
akash1810, line 15:
```suggestion
      val identity = AppIdentity.whoAmI(defaultAppName = "amigo")
```

app/components/AppComponents.scala
```diff
@@ -172,14 +153,14 @@ class AppComponents(context: Context)
     .withClientConfiguration(clientConfiguration)
     .build()
 
-  val amigoUrl: String = configuration.get[Option[String]]("amigo.url").getOrElse(s"https://${identity.app}.gutools.co.uk")
+  val amigoUrl: String = configuration.get[Option[String]]("amigo.url").getOrElse(s"https://amigo.gutools.co.uk")
```
akash1810, line 156: Should this be stage aware?

app/services/ELKLogging.scala
```diff
@@ -6,30 +6,41 @@ import ch.qos.logback.core.Appender
...
+  val identity = appIdentity match {
+    case DevIdentity(_) => None
+    case awsIdentity @ AwsIdentity(_, _, _, _) => Some(awsIdentity)
+  }
+
+  val instanceId = identity.flatMap { _ =>
+    Try(Option(EC2MetadataUtils.getInstanceId)) match {
+      case Success(Some(instanceId)) => Some(instanceId)
+      case _ => None
+    }
```
akash1810, line 31: Could this be:

```scala
Try(EC2MetadataUtils.getInstanceId)).toOption
```

app/services/ELKLogging.scala
```diff
@@ -6,30 +6,41 @@ import ch.qos.logback.core.Appender
...
+
   def getContextTags: Map[String, String] = {
     val effective = Map(
-      "app" -> identity.app,
-      "stage" -> identity.stage,
-      "stack" -> identity.stack,
-      "region" -> identity.region,
+      "app" -> identity.map(_.app).getOrElse(""),
+      "stage" -> identity.map(_.stage).getOrElse("DEV"),
+      "stack" -> identity.map(_.stack).getOrElse("dev-stack"),
```
akash1810, line 40: Are these the correct default values?

app/AppLoader.scala
jacobwinch: Thanks for noticing this 😱

app/components/AppComponents.scala
jacobwinch: `identity.app` was ultimately derived from the `App` tag before, so this has always just been `amigo`

As far as I can see this is only used for notifications anyway and I believe those are `PROD`-only

app/services/ELKLogging.scala
jacobwinch: This is ['borrowed' from `simple-configuration`](https://github.com/guardian/simple-configuration/blob/4c73c166bbe8094dc51fb0364e162af034591af8/core/src/main/scala/com/gu/AppIdentity.scala#L63-L65) - unfortunately it seems like this complex wrapping is necessary.

app/services/ELKLogging.scala
jacobwinch: I don't think it really matters to be honest because identity is only a `None` in dev mode where ELK logging is disabled. We could make them all "unknown" or something like that if you think it helps?

akash1810: "unknown" would be good.

akash1810 (approved):

jacobwinch:
> I will leave the old config in DynamoDB for a while before removing it to allow us to rollback if necessary.

I left this a while longer than I'd intended - but the config that was stored in DynamoDB has now been deleted! 😅

PR #689, opened 2022-01-26
Update test dependencies

jacobwinch:
## What does this change?

This PR updates all test dependencies to ensure that we are using versions which are compatible with the most recent versions of Scala (we are hoping to upgrade soon).

There have been a couple of significant changes since these dependencies were last updated which makes this diff a little noisy (hence the separate PR):

* The library [has been modularised](https://www.scalatest.org/release_notes/3.1.0), which means that we can improve standardisation within the project by only including the [test styles](https://www.scalatest.org/user_guide/selecting_a_style) that we need.
* The library/approach for using ScalaTest with Mockito seems to have changed - I've followed the approach described [here](https://www.scalatest.org/plus/mockito).

## How to test

CI should be sufficient for this PR.

### DIFF
--- a/test/models/CustomisedRoleTest.scala
+++ b/test/models/CustomisedRoleTest.scala
@@ -1,29 +1,31 @@
 package models
 
-import org.scalatest.{ EitherValues, FunSuite, ShouldMatchers }
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
 import cats.syntax.either._
+import org.scalatest.EitherValues
 
-class CustomisedRoleTest extends FunSuite with ShouldMatchers with EitherValues {
+class CustomisedRoleTest extends AnyFlatSpec with Matchers with EitherValues {
 
-  test("should parse a map of variables") {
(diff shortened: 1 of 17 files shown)


test/models/CustomisedRoleTest.scala
```diff
@@ -1,29 +1,31 @@
 package models
 
-import org.scalatest.{ EitherValues, FunSuite, ShouldMatchers }
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
 import cats.syntax.either._
+import org.scalatest.EitherValues
 
-class CustomisedRoleTest extends FunSuite with ShouldMatchers with EitherValues {
+class CustomisedRoleTest extends AnyFlatSpec with Matchers with EitherValues {
```
jacobwinch, line 8: This is an example of the benefits of importing `scalatest-flatspec` rather than `scalatest` - it prevents us from having a single test suite with a different style to all of the others

akash1810 (approved):

PR #690, opened 2022-01-31
Upgrade to Scala 2.12

jacobwinch:
## What does this change?

This PR upgrades the project to use Scala 2.12. Key changes for this codebase include:

* [`Either` becomes right-biased](https://www.scala-lang.org/news/2.12.0/#either-is-now-right-biased), allowing us to remove some boilerplate (and some usages of `cats`)
* [`JavaConversions` is deprecated and replaced by `JavaConverters`](https://www.scala-lang.org/news/2.12.0/#other-changes-and-deprecations), which requires more explicit syntax for conversion between Java and Scala

Although this is not the latest Scala version*, it unblocks a couple of key upgrades:

* Play Framework version 2.8.x, which [drops 2.11 support](https://www.playframework.com/documentation/2.8.x/Migration28#Scala-2.11-support-discontinued)
* Recent versions of [`play-google-auth`](https://github.com/guardian/play-googleauth) (and its mandatory dependency [`play-secret-rotation`](https://github.com/guardian/play-secret-rotation)), which only supports 2.12 and above

Before we can move to Scala 2.13 we need to remove/update several more dependencies, hence the desire to break these upgrades down into multiple stages/PRs.

*As it happens, it's not even the latest version of Scala 2.12. Weirdly, this project will not compile with later versions of 2.11.x _or_ 2.12.x, due to errors related to implicits. I spent a while looking at this but ended up down a rabbit hole, so I think we should use this PR as a stepping stone and revisit this if it's still a problem when we try to move to 2.13 (there are a couple of [answers on Stack Overflow](https://stackoverflow.com/a/66229034/15960596) which give me hope!).

## How to test

I've deployed this to `CODE` and tested some basic functionality (baking an AMI, viewing bake logs, creating/editing recipes).

* We are using a more recent Scala version
* Play Framework / `play-google-auth` upgrades are unblocked

This PR should be pretty low risk given the testing performed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ import java.time.{ZoneId, ZonedDateTime}
 
 name := "amigo"
 version := "1.0-latest"
-scalaVersion := "2.11.8"
+scalaVersion := "2.12.0"
 
 javaOptions in Universal ++= Seq(
   s"-Dpidfile.path=/dev/null",
@@ -74,7 +74,7 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
   "com.gu" %% "scanamo" % "1.0.0-M4",
-  "com.beachape" %% "enumeratum" % "1.3.7",
+  "com.beachape" %% "enumeratum" % "1.6.1",
   "com.typesafe.akka" %% "akka-actor-typed" % "2.5.21",
   "com.gu" %% "simple-configuration-ssm" % "1.5.6",
   "com.typesafe.play" %% "play-iteratees" % "2.6.1",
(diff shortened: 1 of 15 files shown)


build.sbt
```diff
@@ -74,15 +74,15 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
   "com.gu" %% "scanamo" % "1.0.0-M4",
-  "com.beachape" %% "enumeratum" % "1.3.7",
-  "com.typesafe.akka" %% "akka-actor-typed" % "2.5.21",
+  "com.beachape" %% "enumeratum" % "1.6.1",
```
jacobwinch, line 77: The previous `enumeratum` and `fastparse` versions were so old that they were not published for `2.12`.

build.sbt
```diff
@@ -74,15 +74,15 @@ libraryDependencies ++= Seq(
...
+  "com.typesafe.akka" %% "akka-actor-typed" % "2.5.26",
```
jacobwinch, line 78: Play is using `2.5.26` and Akka versions are [supposed to be consistent](https://doc.akka.io/docs/akka/2.5/common/binary-compatibility-rules.html?_ga=2.91349658.1465086831.1643622823-1253382111.1594211568#mixed-versioning-is-not-allowed). Upgrading to `2.12` allows us to fix this.

akash1810 (approved):
🎉

PR #693, opened 2022-02-04
Add play-secret-rotation

jacobwinch:
## What does this change?

This PR adds the [`play-secret-rotation`](https://github.com/guardian/play-secret-rotation) library to this project, which allows us to rotate our [Play Application secret](https://www.playframework.com/documentation/2.8.x/ApplicationSecret) regularly, without impacting users 🎉 

This allows us to improve security and avoid deprecation warnings when upgrading [`play-googleauth`](https://github.com/guardian/play-googleauth).

## How to test

I've deployed this to `CODE` and [tested rotating the Play application secret](https://logs.gutools.co.uk/s/devx/goto/5a2d4fa0-8821-11ec-9387-7946d531de16).

It's easy to rotate our application secret whenever we need to. 

That said, I'm not completely sold on the [final recommendation made by the `play-secret-rotation` library](https://github.com/guardian/play-secret-rotation/blob/main/aws-parameterstore/README.md#secret-updating-lambda) (or at least the suggested implementation of it). It would be nice if this lambda was deployed once per account (or even just once for the organisation) and you could just opt-in to have your secret rotated as part of the `@guardian/cdk` pattern.

I think this is pretty low risk, given the testing performed.

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -16,11 +16,11 @@
     "diff": "cdk diff --path-metadata false --version-reporting false"
   },
   "devDependencies": {
-    "@aws-cdk/assert": "1.132.0",
+    "@aws-cdk/assert": "1.140.0",
     "@guardian/eslint-config-typescript": "^0.6.0",
     "@types/jest": "^26.0.20",
     "@types/node": "15.12.5",
-    "aws-cdk": "1.132.0",
+    "aws-cdk": "1.140.0",
     "eslint": "^7.29.0",
     "jest": "^27.0.6",
     "prettier": "^2.3.2",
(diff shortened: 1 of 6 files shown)


cdk/package.json
```diff
@@ -29,13 +29,13 @@
...
-    "@aws-cdk/aws-s3": "1.132.0",
-    "@aws-cdk/core": "1.132.0",
-    "@guardian/cdk": "31.0.0",
+    "@aws-cdk/aws-ec2": "1.140.0",
+    "@aws-cdk/aws-events-targets": "1.140.0",
+    "@aws-cdk/aws-iam": "1.140.0",
+    "@aws-cdk/aws-lambda": "1.140.0",
+    "@aws-cdk/aws-s3": "1.140.0",
+    "@aws-cdk/core": "1.140.0",
+    "@guardian/cdk": "34.1.0-beta.1",
```
jacobwinch, line 38: We need to use a 'production' version of this library once https://github.com/guardian/cdk/pull/1076 has been released.

akash1810 (approved):
🎉 

Are we able to remove v1 of the AWS SDK now? (well, probably in a different PR).

jacobwinch:
> Are we able to remove v1 of the AWS SDK now? (well, probably in a different PR).

Yep, there is quite a bit of code using v1, but in theory we could refactor it all to use v2 instead. Although it would bring some advantages (not to mention consistency - maybe I should have just used v1 here!), AWS still seem to be releasing/maintaining v1 and it's [acceptable for both libraries to coexist in the same repo](https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-x-released/), so it's probably worth considering whether this is worthwhile if we intend to move away from AMIgo in the near(ish) future.

PR #695, opened 2022-02-09
Upgrade to Play 2.8

jacobwinch:
## What does this change?

This PR upgrades Amigo from Play 2.6 to Play 2.8, based on the following the migration guides:

* [2.6 to 2.7 Migration Guide](https://www.playframework.com/documentation/2.8.x/Migration27)
* [2.7 to 2.8 Migration Guide](https://www.playframework.com/documentation/2.8.x/Migration28)

One recommendation ([updating sbt](https://www.playframework.com/documentation/2.8.x/Migration28#sbt-upgrade)) is not followed yet, as I cannot build the project using the latest version of `sbt`. 

I think we will need to upgrade Scala to unblock this, and that involves updating or removing a number of other dependencies. As `Play 2.8 still supports sbt 0.13`, I think we can decouple those tasks and complete that work in (yet another) upgrade PR!

## How to test

I've deployed this to `CODE` and confirmed that I can still bake an AMI.

We are using the latest version of the Play Framework 🎉 

This is a significant upgrade; we should monitor this service carefully after this change has been deployed in case something was missed during testing.


build.sbt
```diff
@@ -71,16 +71,17 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.dataformat" % "jackson-dataformat-yaml" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
+  "com.fasterxml.jackson.module" %% "jackson-module-scala" % jacksonVersion,
   "com.gu" %% "scanamo" % "1.0.0-M4",
   "com.beachape" %% "enumeratum" % "1.6.1",
-  "com.typesafe.akka" %% "akka-actor-typed" % "2.5.26",
+  "com.typesafe.akka" %% "akka-actor-typed" % "2.6.18",
```
jacobwinch, line 77: Mixing Akka versions is [not allowed](https://doc.akka.io/docs/akka/2.6/common/binary-compatibility-rules.html?_ga=2.118238537.1723156955.1644339084-1253382111.1594211568#mixed-versioning-is-not-allowed), so we upgrade to use [exactly the same version as Play](https://github.com/playframework/playframework/blob/2.8.13/project/Dependencies.scala#L10)

PR #702, opened 2022-03-14
Update teamcity-agent role to work on ARM

nicl:
## What does this change?

Add ARM support to the `teamcity-agent` role so that we can switch to Graviton.

## How to test

Simple local testing seems difficult on an M1 mac (no virtualbox) and having issues running the main app locally too, so testing on CODE for now.

### DIFF
--- a/roles/teamcity-agent/tasks/main.yml
+++ b/roles/teamcity-agent/tasks/main.yml
@@ -95,10 +95,17 @@
 - name: disable systemd as default ddns server
   shell: ln -sf /run/systemd/resolve/resolv.conf /etc/resolv.conf
 
-- name: set default java version # to get available versions run `update-java-alternatives --list`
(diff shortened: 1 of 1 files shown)


philmcmahon:
Is there a risk of breaking builds that don't work on ARM? Though I guess most of our stuff runs on ARM now so this is probably better than the current view of the world

nicl:
@philmcmahon yeah definitely. We're trialling Graviton on tools projects at the moment to get a sense of things. Eventual roll-out strategy is still TBD. We may shift everyone to Graviton and provide an opt-out, or introduce Graviton as optional for a bit to allow people to switch over at a more leisurely pace.

There will be some projects that do not run on ARM so, whichever approach we take, we will support ARM and x86 agent spot fleets for the foreseeable future.

PR #708, opened 2022-03-21
Remove Log streaming from showBake page - second attempt

michaelwmcnamara:
## What does this change?
This PR fixes and reinstates https://github.com/guardian/amigo/pull/700 - "Remove Log streaming from showBake page".
That PR was reverted as we discovered that in removing the log-streaming it also removed a trigger to generate encrypted AMI's should they be requested in the recipe.
The reason this broke is that the trigger used was the occurrence of a "bake completed" message in the logs.

Like its predecessor, this PR removes the streaming for the showBake page - including the log streaming to make an upgrade to scala 2.13 possible.
However, instead of removing the trigger to generate an encrypted AMI, this new PR instead sources the trigger from the Akka actors notifications API, using a BakeEvent to signal completion of the Bake rather than relying on the (now removed) Log streams.

## How to test

Deploy to CODE 
Bake a supported recipe, making sure to add a request for an encrypted AMI to be built - by adding the account id of the accounts you want it built in to the request-encrypted-copy field. 
When the recipe has baked, inspect the  "recent bakes" field of the recipes page: There should be a copy of the baked AM, with "Encrypted tag: true' for each of the AWS accounts you entered in the request-encrypted-copy field.


jacobwinch (approved):
👍

PR #712, opened 2022-04-04
Update sbt

michaelwmcnamara:
## What does this change?

This change updates scala to use version 2.13.
In order to achieve this many dependencies, and plugins have also needed to be updated to compatible versions.
These in turn required code changes to use updated method calls and remove calls to deprecated functions

## How to test
Test by 
- running unit tests locally
- running amigo locally
- deploying amigo to code and running a bake there.

Amigo is running on scala 2.13
Nothing unexpected is happening on any of the recipes baked by this new version of amigo.


app/components/AppComponents.scala
```diff
@@ -115,20 +112,18 @@ class AppComponents(context: Context, identity: AppIdentity)
...
   // initialise logging
   val elkLoggingStream = configuration.get[Option[String]]("elk.loggingStream")
   val elkLogging = new ElkLogging(identity, elkLoggingStream, awsCredsForV1)
 
   implicit val dynamo = {
-    val dynamoClient: AmazonDynamoDB = AmazonDynamoDBClient.builder()
-      .withCredentials(awsCredsForV1)
-      .withRegion(region)
-      .withClientConfiguration(clientConfiguration)
+    val credentialsProvider = StaticCredentialsProvider.create(awsCredsForV2.resolveCredentials())
```
akash1810, line 123: Does `StaticCredentialsProvider` still use the `InstanceProfile`?

app/data/BakeLogs.scala
```diff
@@ -23,42 +25,27 @@ object BakeLogs extends Loggable {
...
-      unprocessedKeys = unprocessedItems.map(_.getDeleteRequest.getKey.asScala.toMap).toList
-      unprocessedKey <- unprocessedKeys
-    } yield unprocessedKey
-
-    if (unprocessedKeys.nonEmpty) {
-      log.warn(s"${unprocessedKeys.size} log entries not processed during deletion, trying again - attempt $attempt")
-      // avoid overwhelming the DB by pausing briefly before mopping up
-      Thread.sleep(BATCH_PAUSE)
-      doDelete(UniqueKeys(unprocessedKeys.toSet), attempt + 1)
-    }
```
akash1810, line 61: Does Scanamo do this recursion now? 😍

app/data/Dynamo.scala
```diff
@@ -76,8 +104,8 @@ class Dynamo(val client: AmazonDynamoDB, stage: String) extends Loggable {
 
   @tailrec
   private def waitForTableToBecomeActive(name: String): Unit = {
-    Try(Option(client.describeTable(name).getTable)).toOption.flatten match {
-      case Some(table) if table.getTableStatus == TableStatus.ACTIVE.toString => ()
+    Try(Option(client.describeTable(generateDescribeTableRequest(name)))).toOption.flatten match {
```
akash1810, line 107: We're moving from `getTable` to `describeTable`, does the instance have the relevant permissions here?

For a future PR: we should create these tables with IaC.

akash1810 (approved):
LGTM! Added a question about IAM permissions. Great work.

nicl (approved):
Looks great!

Something we discussed (and probably merits a Trello card) is to migrate away from the Kinesis Logback Appender. But that is definitely beyond scope here.

app/components/AppComponents.scala
akash1810: Ah I thought `StaticCredentialsProvider` would be API credentials via a key/secret rather than the `InstanceProfile`. Sounds like I'm mistaken though, which is great!

app/data/BakeLogs.scala
michaelwmcnamara: I'm not sure.
As of https://github.com/scanamo/scanamo/pull/486 deleteAll returns Unit. This looks to be because ScanamoFree.deleteAll now uses a "traverse" method from the cats library to run the deletes. According to the comments, this method does not have any way to accumulate results and now returns a value of "void".

If we want to keep this recursion, I could run a for loop across the set of UniqueKeys and use the vanilla "delete" method  on them. This still returns a result. Let me know if I should attempt this.

app/data/Dynamo.scala
michaelwmcnamara: Looking at the cloudformation snapshot, it has wildcard ('*') permissions for any table with a title matching: 
"arn:aws:dynamodb:*:*:table/amigo-PROD-*"
 Definitely worth looking into creating tables with IaC as a future PR.

app/components/AppComponents.scala
nicl: Update, this did indeed break things! Thanks @akash1810 for thorough review and making it easy to debug the issue. The fix (now merged) is here: https://github.com/guardian/amigo/pull/725

PR #724, opened 2022-04-28
Improve cdk base readme

nicl:
Make cdk-base readme less scary and also more informative.

### DIFF
--- a/roles/cdk-base/README.md
+++ b/roles/cdk-base/README.md
@@ -1,26 +1,33 @@
-CDK Base
-========
+# CDK Base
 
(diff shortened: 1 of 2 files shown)


roles/cdk-base/README.md
```diff
@@ -1,27 +1,34 @@
...
 To ship logs, ensure your instance has the following tag:
 
     LogKinesisStreamName
 
 set to the name of your logging Kinesis Stream.
 
+Also ensure your instances have permissions:
+
+    kinesis:DescribeStream
+    kinesis:PutRecord
```
akash1810, line 23: Can we describe which stream these permissions should be for?

PR #726, opened 2022-05-04
Delete roles/mongo-helper-scripts directory

akash1810:
## What does this change?

These are not used.

It looks like they were originally [added](https://github.com/guardian/amigo/pull/87) for flexible-content, which has now [migrated to Postgres](https://www.theguardian.com/info/2018/nov/30/bye-bye-mongo-hello-postgres).


jacobwinch (approved):
Thanks for including the history/reasoning in your PR description! 👍

PR #730, opened 2022-05-09
Deprecate openjdk roles

nicl:
Corretto is the preferred distribution now.

For rationale, see:

https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit#heading=h.bylq0fpat69m


roles/java11/README.md
```diff
@@ -0,0 +1,5 @@
+# Java 8 (openjdk)
```
jacobwinch, line 1:
```suggestion
# Java 11 (openjdk)
```

roles/java11/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+**This role is now DEPRECATED. Switch to the `java11corretto` role. For details
```
jacobwinch, line 3: I think you only need the migration doc if you've come to this role looking to upgrade from `java8`? I.e. if you're already using the `java11` role, presumably you could just swap to the `java11corretto` role without even reading the doc*, because it should be a direct replacement? Perhaps we should cover those two cases in the deprecation notice?

*We could even do this _for_ teams if we wanted to deprecate more aggressively? Obviously we would need to email and let people know that this is going to happen if we decided to do this! I believe AWS did this for their Lambda runtime previously...

jacobwinch (approved):

roles/java17/README.md
```diff
@@ -0,0 +1,3 @@
+# Java 11 (openjdk)
```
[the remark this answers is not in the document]
jacobwinch: As it happens, it doesn't look like anyone is using `java17corretto` yet either, so if we really want to standardise Java 11 (Corretto) then we could remove this for now too?

jacobwinch (approved):

PR #731, opened 2022-05-09
Clearer new & update recipe UI for "request encrypted copy"

kenoir:
## What does this change?

This change updates the new & update recipe UI to describe the "Request encrypted copy" field more thoroughly. When this field has a value the descriptive text is obscured, this change moves the description beneath the field.

This is wha the UI looks like with the update:
Screenshot 2022-05-09 at 14 28 46

## How to test

Application compiles, tests pass!

Reviewers feel the language is clearer and easier to understand to new (or long returning) users of AMIgo.


jacobwinch (approved):
Thanks for this; I agree that this is clearer 👍

PR #733, opened 2022-05-18
[`cdk-base` role] README tweak

twrichards:
When pairing with @akash1810 we decided to make it a little clearer in the README what 'safe' means w.r.t. adding this role to a recipe.


akash1810 (approved):
Thanks!

PR #734, opened 2022-05-18
Revert "[`cdk-base` role] README tweak"

twrichards:
Reverts guardian/amigo#733

Turns out its not 'safe' to use `cdk-base` on recipes which use fluentbit (with their own config) rather than `devx-logging` (see for example https://github.com/guardian/path-manager/pull/52) - that can be investigated in due course (FYI @nicl @jacobwinch @akash1810 )

Reverting this for now.


akash1810 (approved):
😭 thanks for confirming, let's look at this tomorrow.

PR #737, opened 2022-05-31
Upgrade wazuh-agent to v4.3.1-1

louishather:
## What does this change?

This PR changes the role defaults for the wazuh-agent roles, changing the agent_version parameter from 4.1.5-1 to 4.3.1-1. Future wazuh-agents will be installed with the newer version. 

Restart behaviour was also added to make agents capable of registration if registration fails during initial startup. The impact of this is very low, with a frequency of 5 minutes. 

## How to test

- [x] Deploy to AMIgo-CODE and check recipes build
- [x] Check wazuh-agent is v4.3.1-1
- [x] Check wazuh-agent is still able to register with coordinator

Wazuh-agent is installed on every build, so we should make sure this won't introduce any build failures, agent registration failures, or other instability. 

The impact of restarting the service was tested on a micro instance. Without a registration with the wazuh manager, it's extremely minimal, and once registered the service will not restart.

### DIFF
--- a/roles/wazuh-agent/defaults/main.yml
+++ b/roles/wazuh-agent/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 stage: PROD
-agent_version: 4.1.5-1
+agent_version: 4.3.1-1
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -120,11 +120,15 @@
       fi
 
 
-- name: Patch wazuh-agent service to call boot script before starting agent
+- name: Patch wazuh-agent service to call boot script before starting agent and restart on failure
   lineinfile:
-    path: /usr/lib/systemd/system/wazuh-agent.service
+    path: /etc/systemd/system/wazuh-agent.service
     insertbefore: '^ExecStart'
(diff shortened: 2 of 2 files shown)


jorgeazevedo (approved):
I was wondering if it would be good to introduce a maximum number of retries (as explained here https://ma.ttias.be/auto-restart-crashed-service-systemd/) to avoid it infinite looping, but I've actually come down on letting it retry indefinitely. It will give it a sort of... permanent recoverability. If a server-side problem were to cause the agents to exit in error, our whole estate would reconnect once the issue was fixed without having to redeploy.

louishather:
> I was wondering if it would be good to introduce a maximum number of retries (as explained here https://ma.ttias.be/auto-restart-crashed-service-systemd/) to avoid it infinite looping, but I've actually come down on letting it retry indefinitely. It will give it a sort of... permanent recoverability. If a server-side problem were to cause the agents to exit in error, our whole estate would reconnect once the issue was fixed without having to redeploy.

I gave this a bit of thought and came to the same conclusion. The present situation is how you describe - agents that aren't able to register initially will never become registered. 

There's a few other reasons why agents may fail to register beyond server-side problems - security groups need to be configured to allow traffic to Wazuh's registration and reporting addresses on tcp 1514 and 1515, which isn't always the case.

PR #738, opened 2022-06-15
Wazuh-agent role systemd service file fix

louishather:
## What does this change?

#737 introduced an issue where recipes failed to bake, as the path to the systemd service file was incorrect. This was changed between v4.1.5 and v4.3.1. In order for agents to register with the wazuh coordinator, the authentication script should be run before the service starts, which is achieved by inserting "ExecStartPre=/usr/local/bin/authenticate-with-wazuh-manager.sh" into the [Service] section of the service file. 

Previously we inserted a single line into the service file contained in the package. Now, instead of relying on the package version, a new systemd service file will be templated in /etc/systemd/system/wazuh-agent.service.

## How to test

- [x] Deploy on CODE and check builds are successful 

Builds no longer fail due to wazuh-agent lineinfile task not finding a file to edit at /lib/systemd/system/wazuh-agent.service.

### DIFF
--- a/roles/wazuh-agent/tasks/main.yml
+++ b/roles/wazuh-agent/tasks/main.yml
@@ -120,15 +120,10 @@
       fi
 
 
-- name: Patch wazuh-agent service to call boot script before starting agent and restart on failure
-  lineinfile:
-    path: /lib/systemd/system/wazuh-agent.service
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Thanks for fixing this 👍

PR #739, opened 2022-06-22
Migrate Image Copier to CDK

nicl:
Migrate Image Copier stacks to CDK
 
Note, we have not used the GuSnsLambda pattern for now as it does not permit a shared (across accounts) bucket for the lamdba file.

Also note, some of the parameters have been kept for now. This is non-idiomatic in CDK but makes sense here as the stacks are stack-setted and account-level overrides are possible.

Lastly, in the medium term we hope to delete these stacks anyway and share encrypted AMIs directly. See:

https://docs.google.com/document/d/1p0cSup1LmgrqDsnN3HGpk9C10ZDJyTbk4InjWHlhJ_w/edit


imageCopier/cdk/cdk.out/kms-key-stack.template.json
```diff
@@ -0,0 +1,329 @@
+{
```
akash1810, line 1: The `cdk.out` dir is usually gitignored, is there a particular reason we need to include it?

imageCopier/cdk/lib/lambda.ts
```diff
@@ -0,0 +1,174 @@
...
+			functionCodeBucketParam.valueAsString,
+		);
+
+		const kmsKeyArnParam = new CfnParameter(this, 'KmsKeyArn', {
+			description: 'Override the default KMS key if required',
+			type: 'String',
+			default: '',
+		});
+
+		new CfnCondition(this, 'ImportKmsKeyArn', {
```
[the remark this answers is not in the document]
nicl: Thanks. Have now removed most of the parameters but kept the SNS topic ones, as these require the deploy tools account ID, which is sensitive info. (Other solutions discussed here but are more effort and probably not worth it at this point in time.)

nicl:
@akash1810 ready for re-review now I think!

imageCopier/cdk/cdk.out/kms-key-stack.template.json
nicl: (Was an accidental commit.)

imageCopier/cdk/lib/__snapshots__/kms.test.ts.snap
```diff
@@ -0,0 +1,125 @@
...
+      "Value": Object {
+        "Fn::GetAtt": Array [
+          "imagecopierkmskeyB4BF7296",
+          "Arn",
+        ],
+      },
+    },
+  },
+  "Resources": Object {
+    "imagecopierkmskeyB4BF7296": Object {
```
akash1810, line 20: Has the logical ID override taken effect here? I'd have expected this to read "KmsKey"?

akash1810: I think we're missing CI on these changes too. Hopefully it would report a snapshot failure?

nicl:
Update: now tested succesfully e2e on the Deploy Tools account!

akash1810 (approved):

PR #742, opened 2022-07-14
Resolve critical git-url-parse vulnerabiility by bumping cdk from 46.0.0 to 46.0.1

michaelwmcnamara:
## What does this change?
This change updates cdk from 46.0.0 to 46.0.1 - this will resolve a critical vulnerability in the git-url-parse library.
There will be no other impact on the Amigo codebase.

## How to test
Unit tests will provide sufficient confirmation of success.

Snyk vulnerability notification for git-url-parse will no longer be displayed on git reppository.

### DIFF
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -30,8 +30,5 @@
     "ts-node": "^10.0.0",
     "source-map-support": "^0.5.16",
     "typescript": "~4.3.5"
-  },
-  "dependencies": {
-    "yarn": "^1.22.19"
   }
 }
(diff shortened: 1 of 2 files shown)


cdk/package.json
```diff
@@ -22,13 +22,16 @@
...
     "eslint": "^7.29.0",
     "jest": "^27.0.6",
     "prettier": "^2.3.2",
     "ts-jest": "^27.0.3",
     "ts-node": "^10.0.0",
     "source-map-support": "^0.5.16",
     "typescript": "~4.3.5"
+  },
+  "dependencies": {
+    "yarn": "^1.22.19"
```
jacobwinch, line 35: I don't think it's typical to add `yarn` as a dependency here; do you think we need this change?

michaelwmcnamara: Good spot! I'll clear this up now.

jacobwinch (approved):

PR #822, opened 2022-09-01
Update CDK version

nicl:
The immediate motivation is to begin including metadata so that we can understand usage of constructs in the wild.


jfsoul (approved):

akash1810 (approved):
Brilliant!

PR #831, opened 2022-09-07
Kong v1.5

fredex42:
## What does this change?

Adds a Role to deploy Kong 1.5 for CAPI

## How to test

Produce a build with the `Kong v1.5` Role and deploy it

CAPI is able to use updated Kong

Kong is not explicitly marked as "community edition" any more but is still OSS


fredex42:
Note that the reason this does not update the existing `kong` recipe is that we will be running "old" Kong in PROD and "new" Kong in CODE for a while, so we don't want auto-updates to accidentally replace our PROD image with the updated version before we are ready to support it

roles/kong-v1.5/files/kong.logrotate
```diff
@@ -0,0 +1,13 @@
+/usr/local/kong/logs/*.log {
```
nicl, line 1: Not especially familiar with Kong, so just a thought, but is it possible to log to journald via systemd and avoid some of this? https://docs.konghq.com/gateway/latest/plan-and-deploy/systemd/#logging-to-syslog-and-journald.

nicl (commented):
Looks good to me - though I am not familiar with the details of Kong config/startup.

akash1810:
It looks like 1.5.1 was [released in 2020](https://github.com/Kong/kong/releases/tag/1.5.1). From https://docs.konghq.com/gateway/latest/install-and-run/ubuntu/, it looks like the latest version of Kong is 2.8.1.4 ([released in 2022](https://github.com/Kong/kong/releases/tag/2.8.1)). Any reason we're not using the latest version?

fredex42:
@akash1810 It's because 1.5.1 is the only version that gives bidirectional compatibility from v0.14 (which we are currently using in production) to the 2.8 series.  The plan is that we will go to 2.8 (or possibly 3.x if that is released sooner) via 1.5 as a stepping-stone.

roles/kong-v1.5/files/kong.logrotate
fredex42: Yes I'd like to do this, though because this is a large upgrade (v0.14 -> v1.5) I wanted to change the minimum number of things to get it working and once it is proven then improve the setup

fredex42: The current setup does not even start it via systemd(!) but because there might be a reason for this it's on the "incremental improvements" list too

roles/kong-v1_5/files/kong-shim.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
+#
+# Shim for the kong command that adds the --nginx-conf argument.
+# This script's parent directory should be placed on the PATH
+# earlier than /usr/local/bin, where the real kong command resides.
```
JustinPinner, line 5: I can't quite tell if `/usr/local/kong/shims/kong` honours this requirement whereby the shim directory is located before `/usr/local/bin` on the PATH 🤔

fredex42: @JustinPinner Not entirely sure how this works, but it _does_ work:

```bash
root@ip-10-248-78-5:/usr/local/kong# which kong
/usr/local/kong/shims/kong
```

systemd starts it up like this, within the context of root's environment:

```ini
ExecStartPre=/usr/local/bin/kong prepare -p /usr/local/kong
ExecStart=/usr/local/openresty/nginx/sbin/nginx -p /usr/local/kong -c nginx.conf
```

(that is the default installation, which works).  I think that it means you only need the `--nginx-conf` argument for running commands like `kong migrations up`.

fredex42: @JustinPinner Ah, here it is: https://github.com/guardian/amigo/blob/b4004922dfdaebf9141566790836831e87997d7c/roles/kong-v1_5/tasks/main.yml#L38

fredex42: So yes it only applies to admin shells

JustinPinner (dismissed):
This looks good and sensible to me. I'll approve but I don't know if that's sufficient to warrant merging this. Perhaps check in with DevX one last time before merging?

fredex42:
@JustinPinner thankyou very much! @nicl @akash1810 am I good to merge?

roles/kong-v1_5/files/kong-0.14.0.conf
```diff
@@ -0,0 +1,285 @@
+# -----------------------
```
akash1810, line 1: minor: The content of this file feels like a reference implementation. It's a little difficult to see what we're actually configuring due to the volume of comments. Could we simplify by having one comment, that links to Kong's config documentation?

akash1810 (dismissed):
A minor comment, but nothing blocking.

roles/kong-v1_5/files/kong-0.14.0.conf
fredex42: tbh that file can actually be removed, will take it out then merge, thanks!

fredex42: removed that file and placed the actual config settings in an obvious block in the 1.5 configuration with explanatory comments

akash1810 (approved):

PR #873, opened 2022-10-19
Fix for hardened logrotate setup

fredex42:
## What does this change?

Ubuntu 20.04 hardens the logrotate setup, preventing it from writing to most system locations. Unfortuntely, Kong keeps its logs under /usr/local/kong/logs so this must be explicitly allowed in the unit file.

## How to test

Build on CODE and check that the files have the right contents (the `ReadWritePaths` should be added to `logrotate.service`).  Leave it overnight to check that the log files do, in fact, rotate correctly

Reliable log warehousing from Kong

n/a

### DIFF
--- a/roles/kong-v1_5/files/kong.logrotate
+++ b/roles/kong-v1_5/files/kong.logrotate
@@ -5,6 +5,7 @@
 	rotate 4
 	compress
 	delaycompress
+	copytruncate
 	notifempty
 	nocreate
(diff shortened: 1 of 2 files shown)


kenoir (approved):
Looks fine to me.

fredex42:
Tested the resulting build in CAPI CODE and it works :)

PR #903, opened 2022-11-16
Roles for adding Amazon Open Telemetry agent to the box

fredex42:
## What does this change?

Adds two new roles - `otel-collector-agent-arm` and `otel-collector-agent-x86`.  These install Amazon's Open Telemetry agent to the box, which can be used for e.g. Prometheus metric shipping, AWS X-ray tracing, etc.

Further documentation is in the added README files

## How to test

Perform a build with the role enabled.  Once you start an instance with the resulting build, you should see the agent installed as per the README instructions.

You should be able to fire it up by following the README instructions to provide a configuration file and manually start it.

Able to ship metrics

None, this is strictly opt-in.  The software has been tested by DevX and is AWS supported.


roles/otel-collector-agent-arm/README.md
```diff
@@ -0,0 +1,94 @@
+# otel-collector-agent-x86
+
+## What does this do?
+
+It installs the AWS OpenTelemetry agent (https://github.com/aws-observability/aws-otel-collector) onto the target box, without configuration.
```
jacobwinch, line 5: Minor pedantic thing: when implementing the AWS Distro for OpenTelemetry for a JVM-based application there are two main components:

1. The collector (installed here) which ships metrics/traces, usually via a sidecar pattern
2. A [Java agent](https://aws-otel.github.io/docs/getting-started/java-sdk/trace-auto-instr)

In order to disambiguate I think these role names/docs should use the term 'collector' (rather than 'agent') throughout.

roles/otel-collector-agent-arm/README.md
```diff
@@ -0,0 +1,94 @@
...
+    region: "eu-west-1"
+
+receivers:
+  prometheus:
+    config:
+      scrape_configs:
+        - job_name: 'your-app-name'
+          scrape_interval: 60s
+          static_configs:
+            # metrics endpoint for your app
```
jacobwinch, line 41: It might be helpful to point out that it expects metrics to be surfaced via `/metrics` on the specified port, by default?

jacobwinch (approved):
Great work, thanks for adding this! 

I've left a couple of minor comments about naming/docs, but overall this looks great to me 👍

roles/otel-collector-agent-arm/README.md
fredex42: 👍

roles/otel-collector-agent-arm/README.md
fredex42: That makes sense

jacobwinch (approved):
Thanks for updating the docs 👍

PR #904, opened 2022-11-16
Change kong log purging frequency to hourly

fredex42:
## What does this change?

- Update the Kong role to make logrotate run hourly (within a +/- 5min window)
- Update the Kong logrotate rule to also run hourly

This is to alleviate a problem which we believe occurred due to disk pressure on the node. Since Kong is 3rd party software, we cannot directly push logs to Elasticsearch and must read them from disk via Logstash.

## How to test

Check that the build works. Once you have a build, check that the logrotate process is still set up correctly and is now running every hour, round about `:00` minutes.

Improved Kong reliability

This simply modifies existing configurations for logrotate and the resulting configuration has been independently tested to work.  This makes "official" a temporary fix which was put into Kong earlier (https://github.com/guardian/content-api/pull/2714)

### DIFF
--- a/roles/kong-v1_5/files/kong.logrotate
+++ b/roles/kong-v1_5/files/kong.logrotate
@@ -1,5 +1,5 @@
 /usr/local/kong/logs/*.log {
-	daily
+	hourly
 	size 100M
 	missingok
 	rotate 4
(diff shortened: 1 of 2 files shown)


fredex42:
@kenoir It's already tested - well the specific changes to the files are confirmed to do what they should, we did an emergency patch using the instance startup script.

I'm going to do a build from this recipe and then verify that the files look like they should, and then we'll be all good

PR #922, opened 2022-12-05
Upgrade packer to latest

andrew-nowak:
## What does this change?

Upgrade to v1.8.5. Older versions were running afoul of https://github.com/hashicorp/packer/issues/11733 when building images for ubuntu 22.04, eg. https://amigo.gutools.co.uk/recipes/editorial-tools-jammy-java8/bakes/1

This upgrade changed the default location Packer used for its cache directory. Instead of `${PWD}/.packer_cache` it switched to using `${HOME}/.cache/packer`. This shouldn't really have been a problem except that for reasons lost to history the AMIgo box specifies `/home/amigo` as home in `/etc/passwd` but this does not exist and packer was unable to create it.

We fix this by setting `$PACKER_CACHE_DIR` when executing packer and using isolated temp directories which are cleaned up along with the config and ansible files.

## How to test

 - [x] Baking a 22.04 image succeeds in CODE.

Teams can move to the latest Ubuntu LTS.

### DIFF
--- a/app/packer/PackerRunner.scala
+++ b/app/packer/PackerRunner.scala
@@ -4,15 +4,18 @@ import ansible.PlaybookGenerator
 import event.EventBus
 import models.Bake
 import models.packer.PackerVariablesConfig
+import org.apache.http.concurrent.BasicFuture
 import play.api.libs.json.Json
 import services.{AmiMetadataLookup, Loggable, PrismData}
 
-import java.io.File
+import java.io.{File, IOException}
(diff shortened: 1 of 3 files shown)


nicl (commented):
Thanks @andrew-nowak code-wise this looks 👍 . Happy to pair on testing in CODE to confirm it works (ping me to sort a good time) and then we can 🚢 .

jacobwinch (approved):

PR #949, opened 2023-01-05
feat: Connect the ALB to the account's Web ACL

akash1810:
> **Note**
> This change requires an SSM Parameter on path `/account/services/waf/web-acl/primary` to be created first.

## What does this change?

Applies the account's shared [AWS WAF](https://aws.amazon.com/waf/) on the ALB to filter out common exploit requests in line with recommendations from @guardian/infosec.


kenoir (approved):

kenoir (commented):
This may not be the right parameter name if we are creating the Web ACL / WAF config on a per service basis.

akash1810:
In https://github.com/guardian/amigo/pull/961 it's stated that the WAF configuration has been done manually. Closing.

PR #976, opened 2023-01-19
Kong admin port protection

fredex42:
## What does this change?

Adds extra configuration to apply a secondary reverse-proxy for Kong's admin endpoint. The endpoint is still firewall protected in deployment (through the use of Security Groups); this is part of an infrastructure rationalisation and defence-in-depth measure agreed with Infosec.

In addition to the existing Security Group configuration which blocks the port to everything except known SGs, we are adding a basic auth challenge which will be set from Secrets Manager (and ultimately auto-rotated, we hope)

## How to test

Once the instance is built, it will have an extra service called protect-admin-port and associated configuration. The Kong configuration is modified to shift the admin port to 8101, localhost only, and protect-admin-port sets up a simple authenticating nginx proxy to protect it.

The user secret is set at instance start time and will later be rotatable.

Improved backend security in CAPI

### DIFF
--- a/roles/kong-v1_5/files/protect-admin-port/nginx.conf
+++ b/roles/kong-v1_5/files/protect-admin-port/nginx.conf
@@ -16,6 +16,9 @@ http {
     listen 0.0.0.0:8001;
 
     access_log /var/log/protect-admin-port/access.log;
+    location /healthcheck {
+        root /var/www;
+    }
     location / {
       auth_basic "admin";
       auth_basic_user_file "/etc/protect-admin-port/users";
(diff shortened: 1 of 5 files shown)


akash1810:
Could @guardian/content-platforms review this too? 🙏🏽

rowannekabalan (approved):

roles/kong-v1_5/files/protect-admin-port/nginx.conf
```diff
@@ -0,0 +1,28 @@
...
+worker_rlimit_nofile 1048576;
+
+events {
+    worker_connections 1048576;
+    multi_accept on;
+}
+
+http {
+  server {
+    listen 0.0.0.0:8001;
```
akash1810, line 16: This doesn't match the `admin_listen` value in `kong-1.5.1.conf` above. Should it?

fredex42: No. This is the protected port; it's forwarding on to admin_listen

PR #1013, opened 2023-02-09
Add docs that `aws-inspector` is not required for Ubuntu 20

georgeblahblah:
## What does this change?

- Updates the README for the `aws-inspector` role to include that it is not required for images based on Ubuntu 20


kenoir (approved):
👏

PR #1019, opened 2023-02-10
feat: Allow the Data Science team to use AMIgo

akash1810:
## What does this change?
Expand the Google group check to the Data Science group, to enable them to use AMIgo.

An alternative would be to add colleagues into the `departmentGroupId` Google group...

## How to test
- Create the SSM Parameter `/CODE/deploy/amigo/auth.google.dataScientistsGroupId`
- Deploy this branch to CODE
- Check you can still access AMIgo CODE
- Check that colleagues in the data science group can access AMIgo CODE

## TODO before merge
- [ ] Create the SSM Parameter `/PROD/deploy/amigo/auth.google.dataScientistsGroupId`


app/controllers/Login.scala
```diff
@@ -1,31 +1,48 @@
...
+  }
+
+  private def checkGroupMembership()(implicit
+      request: RequestHeader
+  ): Future[Result] = {
+    (for {
+      identity <- checkIdentity()
+
+      // ensure user is in at least one of the Google groups
+      _ = requireGroup[GoogleAuthRequest](includedGroups = googleGroupsToCheck)
```
akash1810, line 40: See https://github.com/guardian/play-googleauth/blob/07f83bd010501b671a4703b013120f37d9911338/play-v27/src/main/scala/com/gu/googleauth/actions.scala#L215-L240 for the implementation.

kenoir (approved):

akash1810:
I've just tested the alternative method, and the test failed! This PR needs a bit more work, as it looks like we're not performing _any_ group check 😱!

app/controllers/Login.scala
```diff
@@ -22,10 +29,53 @@ class Login(
...
+  }
+
+  /*
+  This is a copy of https://github.com/guardian/play-googleauth with one difference: here we check for membership in at least _one_ group, where as play-googleauth checks membership in _all_ groups.
+  TODO update https://github.com/guardian/play-googleauth ?
+   */
+  private def checkGoogleGroupMembership(
+      userIdentity: UserIdentity
+  ): EitherT[Future, Result, Unit] = {
+    groupChecker
```
kenoir, line 53: Can't say I love the complexity of this function and the use of [Cats](https://typelevel.org/cats/) to do what is  a relatively simple check (`userGroups.intersect(googleGroupsToCheck)).nonEmpty`), but as this is copied from the `play-googleauth` library perhaps not an issue to address here.

kenoir (approved):

app/controllers/Login.scala
kenoir: `leftMap` that `subflatMap`!

PR #1024, opened 2023-02-14
Extend ruby-gems role to allow installation of specific version

louishather:
## What does this change?

Previously the ruby-gems role took a single parameter `gems` which is a list of gems to install. This change adds the `version_gems` parameter, which ~~takes a dictionary containing names and specific versions to install~~ <sup>1.</sup> takes a list of strings and splits them on the "=" delimiter:

```version_gems: ['gem-1=1.2.3', 'gem-2=5.2.3']```

Two new tasks have been added to this role; one to install gems with specific versions, and another to fail if no gems are provided as a parameter. The latter task was added to preserve the behaviour of the role failing if `gems` is not provided - the logic is to fail if neither `gems` or `version_gems` are provided. 

The two tasks installing versions are conditional, depending on whether `gems` or `version_gems` were provided, or both. 

## How to test

Deploy on CODE and install gems with both `gems` and `version_gems` parameters, check what happens if the parameters contain common gems, or what happens if the first task installs gems configured in `version_gems` as dependencies of the gems in `gems`. 

- [x] Check installed gems match expected
- [x] Check behaviour when `gems `task installs gems mentioned in `version_gems` as a dependency <sup>2.</sup> 
- [x] Check role fails when neither gems or version_gems are defined 

## What is the value of this?

We needed to get a specific version of `fluentd-plugin-elasticsearch` to match our ElasticSearch cluster. I checked if putting the version in the package name would work, like "fluentd-plugin-elasticsearch==1.2.3" as that works for apt packages, but it fails. 

The behaviour could be to inspect the ```gems``` variable for any strings containing the "=" character, and use them for the version task if there is. That would mirror the behaviour of other roles, but the task definition would be harder to understand. We would also have to check existing uses wouldn't be broken, so I'm going with a separate parameter and task. 

This change shouldn't impact any existing uses of ruby-gems - the existing behaviour of the role has been preserved. 

We'll intentionally not be running with the latest versions of packages, and should consider the potential security implication. This functionality should only be used in specific circumstances where the latest version can't be used. 

## Notes

1. I tried to use a list of dictionaries initially, but AMIgo didn't like that and resulted in a parsing error on the web ui. I've used a list of strings instead. 
2. It was possible to accidentally install multiple versions of the `elasticsearch` gem if `gems` included a gem which needed it as a dependency, i.e. `fluent-plugin-elasticsearch`. The task installing specific versions should come first to avoid this. Confirmed the (now second, latest version) task won't install unnecessary dependencies, or multiple versions.


louishather:
Tested and working

akash1810 (approved):

Issue #1052, opened 2023-03-03
Update teamcity-agent role for 20.04 (or higher)


kenoir:
## Why?

In order to avoid the upcoming transition from Long Term Support (LTS) to Extended Security Maintenance (ESM) and ensure the best security for our EC2 instances, we need to move the TeamCity agents from 18.04 to a more recent version of Ubuntu.

## The problem

Currently trying to bake recipes for `focal-teamcity-agent` fails trying to install docker:

```
[2023-03-03 13:24:49] focal-teamcity-agent: TASK [docker : Install docker] *************************************************
[2023-03-03 13:24:59] focal-teamcity-agent: fatal: [127.0.0.1]: FAILED! => {"cache_update_time": 1677849896, "cache_updated": true, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\" install 'docker-ce=18.06.3~ce~3-0~ubuntu'' failed: E: Version '18.06.3~ce~3-0~ubuntu' for 'docker-ce' was not found
", "rc": 100, "stderr": "E: Version '18.06.3~ce~3-0~ubuntu' for 'docker-ce' was not found
", "stderr_lines": ["E: Version '18.06.3~ce~3-0~ubuntu' for 'docker-ce' was not found"], "stdout": "Reading package lists...
Building dependency tree...
Reading state information...
", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}
```

This is because the `teamcity-agent` role depends on the `docker` role and [sets the `version` variable](https://github.com/guardian/amigo/blob/c450c79feb1aad91fe28dfea8e0f59faea965f05/roles/teamcity-agent/meta/main.yml):

```yml
---
dependencies:
  - role: docker
    version: 18.06.3~ce~3-0
```

This is consumed in [the ansible playbook](https://github.com/guardian/amigo/blob/c450c79feb1aad91fe28dfea8e0f59faea965f05/roles/docker/tasks/main.yml#L33) in this context:

```yml
- name: Install docker
  apt:
    name: docker-ce={{ version }}~ubuntu
    update_cache: yes
    state: present
```

[...]

PR #1056, opened 2023-03-06
Update teamcity-agent role

kenoir:
## What does this change?

This change updates the `teamcity-agent` role so that it can be installed on Ubuntu 20.04 (focal) in order to move TeamCity agents of 18.04 and avoid the upcoming change from LTS to ESM.

See: https://github.com/guardian/amigo/issues/1052

## How to test

Use `multipass` to build a Ubuntu instance as far as is possible (there are some tasks that depend on being run within AWS). As far as possible all code-paths changed have been tested locally. 

Attempt to build a recipe based on 20.04 in AMIgo CODE, can that AMI then be successfully tested in TeamCity?

## What is the value of this?

In order to avoid the upcoming transition from Long Term Support (LTS) to Extended Security Maintenance (ESM) and ensure the best security for our EC2 instances, we need to move the TeamCity agents from 18.04 to a more recent version of Ubuntu.

This change updates the `docker` role, and could break other consumers of that role, however the change has been tested with `multipass` and the same build of `dockerd` is installed on 18.04 and 20.04.


kenoir:
This currently fails with:

```
[2023-03-06 14:42:29] teamcity-agent: TASK [teamcity-agent : set docker to use nexus pull through proxy] *************
[2023-03-06 14:42:29] teamcity-agent: fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "echo '{\"registry-mirrors\":[\"https://nexus.gutools.co.uk:8082\"]}' > /etc/docker/daemon.json", "delta": "0:00:00.002756", "end": "2023-03-06 14:42:29.642293", "msg": "non-zero return code", "rc": 2, "start": "2023-03-06 14:42:29.639537", "stderr": "/bin/sh: 1: cannot create /etc/docker/daemon.json: Directory nonexistent", "stderr_lines": ["/bin/sh: 1: cannot create /etc/docker/daemon.json: Directory nonexistent"], "stdout": "", "stdout_lines": []}
```

jacobwinch (approved):

PR #1080, opened 2023-03-16
out of disk space fix

aaronp:
the Ophan team used to fix their out of diskspace
issue.

Without this setting, the journald logs will be
appended to /var/log/syslog, eventually filling 
up the disk

## What does this change?

This stops journald from appending to  /var/log/syslog, which will (if left unchecked) will fill up and eventually fill up the disk.

## How to test

Good question! I'm just raising this PR to (hopefully) apply the same change as what Ophan did -- though that fix was done to the UserData.

## What is the value of this?

Protection against running out of disk

Additional scrutiny is advised

> **Note**


akash1810:
IIUC this reduces the opportunity for journald logs to fill syslog, however other processes might still cause syslog to expand uncontrollably. Did we consider log rotating syslog?

kenoir (approved):
I've tested this and can see the journald config be updated as expected, and no other issues on instances running based on this AMI.

kenoir:
Linking to related commit on Ophan repo: https://github.com/guardian/ophan/commit/f303e5495f367775580a71a9b46798aed82ab04c

PR #1099, opened 2023-03-28
Browser-testing role should install Chromium on graviton

davidfurey:
## What does this change?

Chrome is not available for arm64 on Ubuntu, so we have decided to use Chromium instead. This PR changes the browser-testing role to use Chromium.

## How to test

Tested in CODE using https://github.com/guardian/editorial-tools-production-monitoring/pull/316

App launches and runs tests as expected.


twrichards (approved):
👏 
Shame the chromium version can't be specified, so we can guarantee compatibility between browser and driver (see https://github.com/guardian/amigo/pull/641)

davidfurey:
Reasonable point. I'm happy to simplify this to always use Chromium, since it will work on both architectures.

akash1810 (approved):
The PR description seems stale, but happy to update that post merge.

PR #1146, opened 2023-04-28
Test amigo loadbalancer with a T4g.Medium instance

michaelwmcnamara:
## What does this change?
This card changes the instance setting for Amigo from t4g.xlarge to t4g.medium.

## Context for this change:
This change implements work defined in [this trello card](https://trello.com/c/13TjUJVi/1946-determine-ideal-size-for-amigo-instance)

This is a follow on from [investigations into a high number of 5XX errors](https://trello.com/c/uchM34sQ/1862-investigate-amigos-low-availability-metric) triggered by the Amigo load balancer.
It was deduced that these errors were generated as a result of the app trying to run on a t4g.small instance as every few minutes the instance would fail its healthcheck and the ASG would terminate and restart the instance.

To confirm increasing resources would resolve the issue, https://github.com/guardian/amigo/pull/1133 was merged to change the instance definition to run on a xlarge size instance for a week. 
[After running for a week on the xlarge server](https://trello.com/c/vjmsJoJ7/1914-on-hold-until-after-26th-april-confirm-that-amigo-has-not-had-any-crashes-while-using-the-larger-t4g-instance), none of the 5XX errors were generated, confirming that increasing the amount of resources fixed the problem.
However Xlarge instances are not cheap. Consequently we want to test to find the cheapest instance size at which stability will be maintained.

This change sets the instance size to Medium as this is the cheapest instance and most likely to throw errors.
After merging this change will be monitored until a week has passed or we start seeing the 5XX errors and failed healthchecks start triggering again.

If the healthchecks start failing for these instances, the next step will be to repeat the process using the t4c.large instance.

## How to test

Merge to code and ensure Amigo builds correctly and runs.
This will guarantee that the code is stable and any problems we encounter are a result of not being able to meet resource demands.

## What is the value of this?

This will enable us to find the cheapest stable instance we can use to run Amigo. The alternative is to leave it running on a t4g.xlarge instance however these are the most expensive instance type.

Potential risks are that we start to encounter instance healthcheck failures and that some scheduled bakes fail.
As soon as we encounter these we will merge a change to increase the size of the amigo instance to Large.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,12 +73,12 @@ scalacOptions ++= Seq(
   "-Xfatal-warnings"
 )
 
-val jacksonVersion = "2.14.2"
+val jacksonVersion = "2.15.0"
 val circeVersion = "0.14.5"
 
 // These can live in the same codebase, see: https://aws.amazon.com/blogs/developer/aws-sdk-for-java-2-x-released/
-val awsV1SdkVersion = "1.12.440"
-val awsV2SdkVersion = "2.20.35"
+val awsV1SdkVersion = "1.12.459"
+val awsV2SdkVersion = "2.20.56"
 
 libraryDependencies ++= Seq(
   ws,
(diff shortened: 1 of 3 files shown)


akash1810 (approved):
The commits look a little confusing, but the net change looks reasonable 👍🏽. Love a binary search!

PR #1188, opened 2023-06-08
Bump java for the image copier

NovemberTang:
Co-author: @tjsilver 
## What does this change?

Java runtime has been bumped to java 8

## How to test

Deploy the stackset to one account. Trigger a bake and verify the image copier runs successfully when the bake completes

## What is the value of this?

We have a department wide SLO to update our runtimes from java 8 to java 11

### DIFF
--- a/cdk/lib/__snapshots__/image-copier-lambda.test.ts.snap
+++ b/cdk/lib/__snapshots__/image-copier-lambda.test.ts.snap
@@ -232,7 +232,7 @@ Object {
             "Arn",
           ],
         },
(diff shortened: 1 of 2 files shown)


nicl (approved):
Non-blocker, but if you are tempted to change 'adopt' to 'corretto' here that would also be a small improvement, as it would align things with the actual AWS runtime:

https://github.com/guardian/amigo/blob/main/.github/workflows/ci.yaml#L31

nicl (approved):

PR #1193, opened 2023-06-12
Improve Image Copier docs

nicl:
## What does this change?

This proved challenging to deploy recently, so updating the docs to help here.

cc @NovemberTang and @tjsilver who did this recently and may have thoughts on how this could be improved.

## How to test

n/a - docs change, though someone could step through it if keen.

## What is the value of this?

Easier deployment in future.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -36,8 +36,9 @@ new ImageCopierKMSKey(app, "kms-key-stack", {
   description: "AMIgo kms key creator",
 });
 
-new ImageCopierLambda(app, "lambda-stack", {
+new ImageCopierLambda(app, "imagecopier-lambda-stack", {
   stack: "deploy",
   stage: "PROD",
   description: "AMIgo image copier lambda",
+  version: "v1", // Update this when the lambda is updated.
 });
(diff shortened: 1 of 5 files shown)


akash1810 (commented):
Should we script the deployment to each stage, allowing deployment to be simply:

```sh
yarn deploy --CODE
```

(or something similar).

https://github.com/guardian/aws-cost-management/pull/14 to offer an example. Note `aws-cost-management` implements StackSets slightly differently from this repository as:
  - It deploys to an OU, rather than account IDs
  - It is defined as a resource in a CloudFormation stack

nicl:
(Copying in non-Github conversation with @akash1810 .)

CDK Deploy is a bit trickier in this case as it has a lambda, specifically a Scala one, which wouldn't integrate so easily.

I'm also concerned that we'd be spending time on a non-RR solution here when it would be better to think about a RR approach going forward.

Instead, I think we should focus on:

1) removing the need for these lambdas - see https://trello.com/c/NQfqV6wS/2041-remove-image-copier-lambdas
2) figuring out how to RRify stacksets (or stop using them)

I'd much rather spend time on either/both the above then further changes to this repo to be honest as it feels like the benefits would be larger.

akash1810 (approved):

PR #1200, opened 2023-06-15
Kong recipe upgrades

fredex42:
## What does this change?

Makes the Kong setup more generic so that the version and architecture can be overridden by arguments.

> **Note**


JustinPinner (commented):
This looks reasonable I think 👍

JustinPinner (approved):
Sorry, that last comment was supposed to be an approval 🙄

PR #1231, opened 2023-07-21
`aws-tools` Role Uses `awscli` v1.29.8

JamieB-gu:
This is the current latest version of the awscli.

Attempts to solve a problem where the awscli was failing to install correctly. The problem was fixed in a recent version of the awscli: https://github.com/aws/aws-cli/issues/8036#issuecomment-1640932993

The awscli was previously pinned to a specific version due to a lack of support for Python 3.4/3.5: https://github.com/guardian/amigo/pull/549

Alternative to #1232


JamieB-gu:
Closing in favour of #1232.

PR #1281, opened 2023-09-06
add ghostscript role to install version from source

marjisound:
## What does this change?
Co-authored with @philmcmahon 

This PR adds a new role for `ghostscript` in order to download ghostscript from source and install it. 

## How to test
This was tested locally using multipass. 

## What is the value of this?
The current ubuntu versions in amigo, do not provide the latest version of ghostscript. So using this role we can install any available version of the ghostscript.

> **Note**


philmcmahon:
This role was succesfully tested here https://amigo.code.dev-gutools.co.uk/recipes/jammy-ARM - it is _very_ slow but that is probably just because it's doing a large compilation step on a slow instance. As it does complete succesfully and the role is optional I think this would be great to get merged so the investigations team can use the most recent version of ghostscript.

roles/ghostscript/README.md
```diff
@@ -0,0 +1,10 @@
+# Ghostscript
+
+Downloads the Ghostscript source from https://github.com/ArtifexSoftware/ghostpdl-downloads/releases
+
+By default it downloads gs10012/ghostscript-10.01.2.tar.gz version but this can be configured to any other existing version.
+
+Then, it unzip the package and installs it on the machine.
```
philmcmahon, line 7: Maybe add a warning about this role being slow (looks like 35mins or so) here?

akash1810 (approved):
Not tested myself, but [this comment](https://github.com/guardian/amigo/pull/1281#issuecomment-1711497032) suggests a successful test has previously been performed.

PR #1313, opened 2023-10-10
Use HTTPS S3 urls to download CloudWatchMonitoring scripts

davidfurey:
## What does this change?

This changes is motivated by Editorial tools bakes failing because HTTP endpoints now return 403 for this resource.

But HTTPS is A Good Thing, so we probably should have been doing this anyway.

I haven't seen any notification from AWS of this change.

## How to test

Minor change, testing almost not required.  But triggering a bake of any of the recently failed ed tools recipes should verify the fix.


jacobwinch (approved):
Thanks @davidfurey!

PR #1314, opened 2023-10-12
Remove VPN required hint from failed bake notification

georgeblahblah:
As part of the response to the incident, AMIgo is now available without the VPN.


akash1810 (approved):
Thanks!

PR #1315, opened 2023-10-12
add formatting detail to docker role readme

ellenmuller:
## What does this change?

This adds a little extra detail in the README about what to pass into the docker version parameter when using docker as a role, to hopefully prevent other people from messing with it for as long as I've done...


AshCorr (approved):
👍

PR #1326, opened 2023-10-25
Add role to install Whisper.cpp

zekehuntergreen:
Co-Authored-By: Philip McMahon <philip.mcmahon@guardian.co.uk>

## What does this change?
Adds a new role `whisper-cpp` which
- clones [whisper.cpp](https://github.com/ggerganov/whisper.cpp) - a "high-performance inference of [OpenAI's Whisper](https://github.com/openai/whisper) automatic speech recognition (ASR) model"
- downloads model `large`
- compiles whisper

## How to test
- tested locally with multipass
- We're testing in CODE by adding the new role to an existing AMI and doing a [successful bake](https://amigo.code.dev-gutools.co.uk/recipes/investigations-ocrmypdf/bakes/5)

## What is the value of this?
In https://github.com/guardian/giant/pull/141 the investigations team is adding a transcription extractor to Giant which will run whisper on audio files. 
We'll need to add the new `whisper-cpp` role to the AMI used for Giant workers in order to make this possible.

We're cloning v1.4.0 of whisper.cpp (the most recent release as of writing) rather than main in order to avoid installing potentially broken future versions.

> **Note**


philmcmahon (approved):
LGTM 

WHISPER

PR #1341, opened 2023-11-13
Whisper.cpp: remove large model, parameterise git commit id to work from

philmcmahon:
## What does this change?
Our whisper.cpp build timed out. This PR aims to help the situation by:
 - removing the 'large' model (over 3 GB) - it's unlikely we'll be able to use this model in production any time soon
 - not throwing away the output from the downloads (so we can see where they fail)

In addition, given that new features to whisper.cpp are being added a lot faster than they are doing releases, I've changed our checkout behaviour so that we fetch the whole repo then reset to  a specific commit ID. This way we can either pin to a certain release or jump to the latest if we see something promising in recent changes to whisper.cpp 

This needs a quick test on amigo CODE prior to merging

### DIFF
--- a/roles/whisper-cpp/defaults/main.yml
+++ b/roles/whisper-cpp/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
 target_dir: /opt/whisper
+whisper_cpp_commit_hash: 9f8bbd3feee603f9ada479fa9c39c03acf2cc6e1
--- a/roles/whisper-cpp/tasks/main.yml
+++ b/roles/whisper-cpp/tasks/main.yml
@@ -11,12 +11,13 @@
     mkdir -p {{ target_dir }}
     cd {{ target_dir }}
     echo "cloning"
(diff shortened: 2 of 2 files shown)


roles/whisper-cpp/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 target_dir: /opt/whisper
+whisper_cpp_commit_hash: 9f8bbd3feee603f9ada479fa9c39c03acf2cc6e1
```
zekehuntergreen, line 3: non blocking: I'm wondering if it would be worth limiting the version we check out to a release tag in order to reduce the risk of deploying an unstable version

philmcmahon: I'm not sure - I think maybe just a commit hash gives us more flexibility as e.g. the quantized models work on `main` now but there hasn't been a new release

philmcmahon:
This is working - most recent build here used it https://amigo.code.dev-gutools.co.uk/recipes/investigations-ocrmypdf

zekehuntergreen (approved):

PR #1373, opened 2023-12-15
fix!: Make `configure-aws-kinesis-agent` fail fast

akash1810:
## What does this change?
Add `set -e` to the configure script means it'll exit at the first non-zero exit code. That is, it fails fast.

This means if we're unable to create the configuration JSON file, then we fail, and don't attempt to start the systemd process.

## How to test
TBD.

## What is the value of this?
Improved DX.

Currently, if the JSON file building fails, we still start the service. However, as the service has bad config, it won't do anything useful.

This change should make it easier to triage issues, as we'll see the service didn't even start.

TBD.

> **Note**


akash1810:
Closing as we're moving to adoption of https://github.com/guardian/devx-logs more and more now. We should see the `aws-kinesis-agent` role b removed in the near future.

PR #1378, opened 2023-12-20
(whisper cpp role) Use parameter to determine which whisper models are downloaded

philmcmahon:
## What does this change?
We're doing a lot of experimentation with different whisper models at the moment and having to modify the role every time is getting a bit tedious. This change parameterises the names of the models which are downloaded when baking an AMI with the whisper role.

## How to test

There's a bake currently in progress [here](https://amigo.code.dev-gutools.co.uk/recipes/investigations-ocrmypdf/bakes/13) - let's see if it works....

### DIFF
--- a/roles/whisper-cpp/tasks/main.yml
+++ b/roles/whisper-cpp/tasks/main.yml
@@ -5,12 +5,13 @@
     - build-essential
     state: present
 
-- name: Clone and compile whisper.cpp
+- name: Clone whisper, clone and compile whisper.cpp
   shell: |
     set -e
     mkdir -p {{ target_dir }}
     cd {{ target_dir }}
     echo "cloning"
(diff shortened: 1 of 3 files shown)


roles/whisper-cpp/tasks/main.yml
```diff
@@ -14,15 +14,24 @@
...
-    cd {{ target_dir }}
+    exit 0
+
+- name: Download models and set permissions
+  shell: |
+    cd {{ target_dir }}/whisper.cpp
+    bash models/download-ggml-model.sh {{ item }}
     chown ubuntu ./ --recursive
     exit 0
+  loop: "{{ whisper_models }}"
```
zekehuntergreen, line 27: nice, I didn't know you could do this!

zekehuntergreen (approved):

PR #1389, opened 2024-01-10
Upgrade all dependencies to Play 3

AshCorr:
## What does this change?

Updates all of our dependencies to versions that use Play 3. Right now we have both Akka and Pekko in our dependency tree which isn't ideal.

Before:

image

After:

image


adamnfish (approved):
LGTM, assuming it starts up and runs ok 👍

PR #1501, opened 2024-07-05
update aws_rds_root_cert_ca role to load new certificate bundle

pvighi:
## What does this change?
This PR updates the aws_rds_root_cert_ca to support new rds certificates which come in bundles instead of single files.

Looking at the [documentation on using ssl in RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html) it looks like now instead of a single certificate there's bundles. The current code assumes the file that it gets from AWS contains a single certificate so if we try to use the new urls it would only load the first one.

A good side effect of loading the certificate in bundles is that it includes the deprecated certificate as well so applications that use it would work with either, which makes migrating with minimal downtime easier.

## How to test
 
For a first test we can sign in to an instance that uses this role and find the generated trust store file (by default `/usr/lib/ssl/certs/java/aws-rds-cacerts`) and list the certificates in it with something like  the keytool cli command `keytool -list -keystore ./[FILE_NAME_HERE]-rfc. This should list the certificates from the selected bundle.

Of course the real test is to just redeploy an instance using this role and being able to connect to rds database using any of the certificates in the bundle.

## What is the value of this?

RDS is deprecating a certificate soon which is still in use in databases, so we need to change it.

### DIFF
--- a/roles/aws-rds-root-cert-ca/tasks/main.yml
+++ b/roles/aws-rds-root-cert-ca/tasks/main.yml
@@ -1,12 +1,20 @@
 ---
-- name: get aws RDS root CA certificate
+- name: get RDS CA certificate bundle
   get_url: url={{certificate_source}}{{certificate_name}}.pem dest=/tmp/{{certificate_name}}.pem
 
-- name: convert pem to der format
-  command: openssl x509 -outform der -in /tmp/{{certificate_name}}.pem -out /tmp/{{certificate_name}}.der
+- name: create directory for cert files
+  file:
+    path: /tmp/rds_certs/
+    state: directory
+- name: split bundle into individual files
+  command: csplit -sz /tmp/{{certificate_name}}.pem  -f /tmp/rds_certs/rds_cert_ '/-BEGIN CERTIFICATE-/' '{*}'
 
-- name: create new truststore
-  command: keytool -import -alias rds-root -keystore "{{truststore_dir}}{{truststore_name}}" -file /tmp/{{certificate_name}}.der -storepass "{{aws_rds_truststore_password | mandatory}}" -noprompt
+- name: add certificates to a new trust store
+  shell: |
+      for cert in /tmp/rds_certs/rds_cert_*; do
+          keytool -importcert -file $cert -alias $cert -keystore {{truststore_dir}}{{truststore_name}} -storepass "{{aws_rds_truststore_password | mandatory}}" -noprompt
+          rm $cert
+          done
 
-- name: import default truststore into new one
+- name: import default trust store into the new one
   command: keytool -importkeystore -srckeystore "{{truststore_dir}}cacerts" -srcstorepass "{{castore_password}}" -destkeystore "{{truststore_dir}}{{truststore_name}}" -deststorepass "{{aws_rds_truststore_password | mandatory}}"
--- a/roles/aws-rds-root-cert-ca/vars/main.yml
+++ b/roles/aws-rds-root-cert-ca/vars/main.yml
@@ -1,5 +1,5 @@
 ---
-certificate_source: https://s3.amazonaws.com/rds-downloads/
-certificate_name: rds-ca-2019-root
+certificate_source: https://truststore.pki.rds.amazonaws.com/global/
+certificate_name: global-bundle
 truststore_dir: /usr/lib/ssl/certs/java/
 truststore_name: aws-rds-cacerts


roles/aws-rds-root-cert-ca/tasks/main.yml
```diff
@@ -1,12 +1,20 @@
...
+    state: directory
+- name: split bundle into individual files
+  command: csplit -sz /tmp/{{certificate_name}}.pem  -f /tmp/rds_certs/rds_cert_ '/-BEGIN CERTIFICATE-/' '{*}'
 
-- name: create new truststore
-  command: keytool -import -alias rds-root -keystore "{{truststore_dir}}{{truststore_name}}" -file /tmp/{{certificate_name}}.der -storepass "{{aws_rds_truststore_password | mandatory}}" -noprompt
+- name: add certificates to a new trust store
+  shell: |
+      for cert in /tmp/rds_certs/rds_cert_*; do
+          keytool -importcert -file $cert -alias $cert -keystore {{truststore_dir}}{{truststore_name}} -storepass "{{aws_rds_truststore_password | mandatory}}" -noprompt
```
pvighi, line 15: the alias for the certificates will just be just the name of the files they were split into ( rds_cert_[NUMBER]).

Unfortunately the file aws provide doesn't contain any labels on the certificates so if we wanted to use a more meaningful alias we would have to extract it from the certificate itself adding a step before the import that does something like `openssl x509 -noout -subject -in $cert`,  extracting the CN section there and making sure it's a valid format for an alias. 
I'm not sure if it's worth the effort to try though as the alias doesn't seem to matter

roles/aws-rds-root-cert-ca/vars/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-certificate_source: https://s3.amazonaws.com/rds-downloads/
-certificate_name: rds-ca-2019-root
+certificate_source: https://truststore.pki.rds.amazonaws.com/global/
+certificate_name: global-bundle
```
pvighi, line 3: This is a bundle of all certificate types for all regions, optionally this can be configured to load certificates for only one region.  Should we  just use the  Dublin certificates? 

The documentation linked in the description mentions this about the bundle:
```
The bundle contains both the rds-ca-2019 intermediate and root certificates. 
The bundle also contains the rds-ca-rsa2048-g1, rds-ca-rsa4096-g1, and rds-ca-ecc384-g1 root CA 
certificates. Your application trust store only needs to register the root CA certificate.
```
I'm not sure if it's any harm in including the certificate that we "don't need" . I don't see a bundle with only the root certificates anywhere

akinsola-guardian (approved):
LGTM

roles/aws-rds-root-cert-ca/vars/main.yml
adamnfish: I think for this amigo role, it makes sense to use the global bundle. We can't be sure here which regions the end users might be using, and making this a parameter couples the role to the implementation detail of the region we're deploying to. If want to make this more specific I'd consider recommending teams move this further along their infrastructure pipelines, e.g. to userdata, so that they know the actual region at that point.

That said, you're the only users of this role as far as I can tell, so whatever works best for you!

roles/aws-rds-root-cert-ca/tasks/main.yml
adamnfish: Would it be easier to try to import the ca bundle as-is, rather than splitting it into certificates and importing each of those separately?

Alternatively, we could take [the approach service-catalogue uses](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesAllRegions), and import this ca into the OS, without using keytool. I'm not 100% sure this works though, if Java has its own separate certificate store?

Perhaps you have considered one or both of these already, and in any cae if you're comfortable with how this PR works now then no worries 👍

pvighi: TI tested just changing the variables on the role and that didn't work with the code in` main`. but now that I changed the parameters to keytool it might work to load the whole bundle in one go.  I'll test it

AshCorr (approved):
LGTM!

roles/aws-rds-root-cert-ca/tasks/main.yml
pvighi: It looks like it still not supported to just load the whole bundle in one go with the keytool.
the aws documentation has an [example script](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL-certificate-rotation.html#UsingWithRDS.SSL-certificate-rotation-sample-script) that is very similar to this (except that they have some extra code to extract more descriptive aliases )

PR #1505, opened 2024-07-08
Add logging for user login and for recipe creation, edit and deletion

marjisound:
## What does this change?

Adding info log messages for the following activities:
- recipe creation
- recipe editing
- recipe deletion
- user logging in

## How to test

Tested this in CODE environment 
## What is the value of this?

We realised that one of our recipes was deleted but couldn't find any detail in the logs when/how that happened. It would be useful to have logs for these scenario. 

> **Note**

### DIFF
--- a/app/controllers/Login.scala
+++ b/app/controllers/Login.scala
@@ -3,7 +3,12 @@ package controllers
 import cats.data.EitherT
 import cats.instances.future._
 import cats.syntax.applicativeError._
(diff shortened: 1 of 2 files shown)


philmcmahon (approved):
This looks like a good security/auditing improvement to me - maybe worth someone from DevX checking they're happy as well

akash1810 (approved):

PR #1507, opened 2024-07-11
Add launch template support to AMIgo

philmcmahon:
## What does this change?
We encountered an issue recently where AMIgo was reporting a recipe as unused, when it actually was in use but via a [launch template](https://docs.aws.amazon.com/autoscaling/ec2/userguide/launch-templates.html) rather than a [launch configuration](https://docs.aws.amazon.com/autoscaling/ec2/userguide/launch-configurations.html).

This PR should prevent that from happening again by adding launch template support to AMIgo - so that when it checks whether a bake is in use, it will also check for usages within launch templates.

## How to test
I tested this on AMIgo CODE against Prism CODE (with this branch https://github.com/guardian/prism/pull/918) by adding an AMIgo CODE AMI to a launch template and checking that the usage is picked up by AMIgo

### DIFF
--- a/app/prism/RecipeUsage.scala
+++ b/app/prism/RecipeUsage.scala
@@ -3,7 +3,7 @@ package prism
 import data.{Bakes, Dynamo, PackageList, Recipes}
 import models.{AmiId, Bake, BakeId, Recipe, RecipeId}
 import play.api.libs.json.Json
-import prism.Prism.{Image, Instance, LaunchConfiguration}
+import prism.Prism.{Image, Instance, LaunchConfiguration, LaunchTemplate}
 import services.PrismData
 import play.api.libs.json.OWrites
 
@@ -14,7 +14,8 @@ case class BakeUsage(
     bake: Bake,
     viaCopy: Option[Image],
     instances: Seq[Instance],
-    launchConfigurations: Seq[LaunchConfiguration]
+    launchConfigurations: Seq[LaunchConfiguration],
+    launchTemplates: Seq[LaunchTemplate]
 )
 
 case class SimpleBakeUsage(bakeId: BakeId, packageListS3Location: String)
@@ -49,6 +50,7 @@ object SimpleBakeUsage {
 case class RecipeUsage(
     instances: Seq[Instance],
     launchConfigurations: Seq[LaunchConfiguration],
+    launchTemplates: Seq[LaunchTemplate],
     bakeUsage: Seq[BakeUsage]
 )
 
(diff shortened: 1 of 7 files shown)


app/prism/RecipeUsage.scala
```diff
@@ -87,8 +90,12 @@ object RecipeUsage {
     val launchConfigurations = prismAgents.allLaunchConfigurations.filter(lc =>
       amiIds.contains(lc.imageId)
     )
+    val launchTemplates =
+      prismAgents.allLaunchTemplates.filter(lt => amiIds.contains(lt.imageId))
     val amis =
-      (instances.map(_.imageId) ++ launchConfigurations.map(_.imageId)).distinct
+      (instances.map(_.imageId) ++ launchConfigurations.map(
```
adamnfish, line 96: I think with launch templates being versioned, that there's a bit more to think about here. At the moment this is brushed under the rug a little by having prism only index the most recent launch template, but we'd likely need something a bit more sophisticated that considers:

- whether any old versions are still in use
- how far back we'd like launch templates to remain deployable

On the latter point, this might consider a time threshold as well as a number of versions (e.g. to only support recent versions, while also ensuring that the previous version is always deployable for rollback).

This is a sticky point, let's chat about it!

philmcmahon: Discussed in this PR https://github.com/guardian/prism/pull/918 - I opted to stick with the limited version of the feature where if an AMI is not in use by a launch template version that is in use by an ASG then we can delete that template

test/housekeeping/MarkOldUnusedBakesForDeletionSpec.scala
```diff
@@ -79,14 +79,14 @@ class MarkOldUnusedBakesForDeletionSpec extends AnyFlatSpec with Matchers {
   def getBakes(recipeId: RecipeId): Iterable[Bake] =
     Iterable(oldBake1, oldBake2, newBake1, newBake2)
   def getEmptyRecipeUsage(bakes: Iterable[Bake]): RecipeUsage =
-    RecipeUsage(Seq.empty, Seq.empty, Seq.empty)
+    RecipeUsage(Seq.empty, Seq.empty, Seq.empty, Seq.empty)
```
adamnfish, line 82: I have a lot of comments about the general state of the tests here, but the most pressing and relevat is that we seem to only be testing for the presence of a BakeUsage instance, rather than checking that BakeUsage actually has any usages in it. This is super confusing here, where it feels like we would be wanting to add a test to check that launch templates are considered, but haven't.

This is all to say, I was expecting to see another test but having dug in I think I can see why that isn't necessary. This definitely feels like an area we could tighten things up, but I'm calling this out for my own benefit rather than expecting any changes here!

adamnfish (approved):
The change looks ok to me, without being an expert on the codebase! I am assuming you've actually checked this works / we'll give it some attention when it's out there, because I haven't considered any implications in the codebase beyond this diff.

LGTM!

adamnfish (approved):
As before

PR #1513, opened 2024-07-15
Add required packages

Georges-GNM:
## What does this change?

Related to this [prodmon PR](https://github.com/guardian/editorial-tools-production-monitoring/pull/337) and enables it to work, essentially - otherwise this will lead to an `UnsatisfiedLinkError`: 
image

I haven't dug deeply, but from what I can tell these used to get installed as part of the java 8 role, but for some reason don't anymore after switching to java 11 - which actually can be confirmed from a couple of bakes of the [browser-testing-jammy-ARM-WITH-cdk-base-java11](https://amigo.gutools.co.uk/recipes/browser-testing-jammy-ARM-WITH-cdk-base-java11): this [first one](https://amigo.gutools.co.uk/recipes/browser-testing-jammy-ARM-WITH-cdk-base-java11/bakes/1) was accidentally still set to use java 8, but when I baked a [second one](https://amigo.gutools.co.uk/recipes/browser-testing-jammy-ARM-WITH-cdk-base-java11/bakes/2) correctly using java 11, the package changes log shows those have not been installed. 

From what I can tell they are required to enable selenium/chromedriver to work, so it feels reasonable to include them in the role directly. 

## How to test

The prodmon PR has temporarily been set to use an AMI baked in code with this role - deploy that to prodmon secondary, and the app should keep working as expected. 

> **Note**


davidfurey:
I suspect that this is due to the change from `openjdk-8-jdk` to `java-11-amazon-corretto-jdk`
`openjdk-8-jdk` is a non-headless version of the JDK, and includes a dependency on `libx11-6` and `libxt-dev` which are both graphics related libraries, which probably (unchecked) eventually depend on the missing libraries.

I assume `java-11-amazon-corretto-jdk` is a headless version of the JDK. The AWS docs don't make this clear, and I haven't downloaded the `.deb` to check.

It might be more robust to switch to `openjdk-11-jdk` for this project rather than manually selecting this specific packages.

Georges-GNM:
> I suspect that this is due to the change from `openjdk-8-jdk` to `java-11-amazon-corretto-jdk` `openjdk-8-jdk` is a non-headless version of the JDK, and includes a dependency on `libx11-6` and `libxt-dev` which are both graphics related libraries, which probably (unchecked) eventually depend on the missing libraries.
> 
> I assume `java-11-amazon-corretto-jdk` is a headless version of the JDK. The AWS docs don't make this clear, and I haven't downloaded the `.deb` to check.
> 

That would make sense, although I thought there was a headless variant rather than it being the default... In any case, the missing libraries do seem to indicate that. 

> It might be more robust to switch to `openjdk-11-jdk` for this project rather than manually selecting this specific packages.

Will give that a try!

akash1810:
> Will give that a try!

Marking this PR as draft, to avoid this being included in the daily message to the team via https://github.com/guardian/actions-prnouncer. Please undo this if I'm mistaken!

Georges-GNM:
Closing as we've preferred an alternative solution (using a different version of java) than adding these packages.

Georges-GNM:
Reopening as this now seems preferable to using an openjdk version of java. 

> It might be more robust to switch to openjdk-11-jdk for this project rather than manually selecting this specific packages.

Appreciate this, but trying to use openjdk 17 for the purpose of browser testing felt like it was displaying some unexpected behaviour on the instance (e.g. re cpu utilisation): 
image

roles/browser-testing/tasks/main.yml
```diff
@@ -14,3 +14,12 @@
 
 - name: Enable Xvfb service
   service: name=xvfb enabled=yes
+
+- name: Install required libxrender1 package
```
davidfurey, line 18:
```suggestion
- name: Install libxrender1, required by Selenium
```

roles/browser-testing/tasks/main.yml
```diff
@@ -14,3 +14,12 @@
...
+  apt: name=libxrender1 state=latest
+
+- name: Install required libxtst6 package
```
davidfurey, line 21:
```suggestion
- name: Install libxtst6, required by Selenium
```

roles/browser-testing/tasks/main.yml
```diff
@@ -14,3 +14,12 @@
...
+  apt: name=libxtst6 state=latest
+
+- name: Install required libxi6 package
```
davidfurey, line 24:
```suggestion
- name: Install libxi6, required by Selenium
```

davidfurey (dismissed):
Looks good to me

davidfurey (approved):

PR #1514, opened 2024-07-16
the url for downloading Kong has moved

paulmr:
## What does this change?

As part of our `kong` [role](https://amigo.gutools.co.uk/roles#kong), we download the application [Kong](https://konghq.com/) from a known URL (this is an essential application that implements the API keys for CAPI).

However, a few months ago, this stopped working. A bit of digging revealed that the Kong team had changed the URL for the download.

I have updated the URL for the role, and I have tested that it now works by first running the role locally with `multipass` and then deploying this branch to code and [re-running the build](https://amigo.code.dev-gutools.co.uk/recipes/kong-v3-arm/bakes/2). I also confirmed in the former case that if I ssh onto the box I can see that kong is installed.

I have also [deployed](https://riffraff.gutools.co.uk/deployment/view/3171679b-7eae-4f84-85b3-dcb6267ea19c) a [branch](https://github.com/guardian/content-api/tree/pmr/tmp-deploy-code-ami) to Kong `CODE` which uses the AMI from the `CODE` Amigo recipe above and I have confirmed that Kong is still working:

image

image

I am also updating some of the default params to better match what we actually use.

## How to test

Build the image, make sure it works and installs Kong OK.

## What is the value of this?

Our AMI for Kong has not been updated in over 40 days because the build has been failing.

We are the only consumers of this role and it currently isn't working at all.

> **Note**

### DIFF
--- a/roles/kong/defaults/main.yml
+++ b/roles/kong/defaults/main.yml
@@ -1,6 +1,6 @@
 ---
-kong_version: 1.5.1
+kong_version: 3.1.1
 kong_user: content-api
 postgres_version: 12
-kong_arch: amd64
-kong_repo: gateway-3.x-ubuntu-focal
+kong_arch: arm64
+kong_repo: gateway-31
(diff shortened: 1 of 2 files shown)


fredex42 (approved):
😎

PR #1562, opened 2024-10-07
feat(aws-tools)!: Update AWS CLI to v2

akash1810:
## What does this change?
Install the latest version of the AWS CLI (v2) using an AWS maintained mechanism.

It looks like installing the AWS CLI via `pip3` isn't maintained by AWS as:
- It (still) installs v1
- It is not listed on https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html

Install the AWS CLI via an AWS provided zip file instead.

The package gets installed to `/usr/local/bin/aws`, replicating existing behaviour. This should make this change backwards compatible for places where we're [using the CLI via its full path](https://github.com/search?q=org%3Aguardian+%2Fusr%2Flocal%2Fbin%2Faws++NOT+is%3Aarchived&type=code).

<details><summary>Why not snap?</summary>
<p>

I originally intended to install the CLI via the official `snap` package. However this'll install to `/snap/bin/aws`, and we'd have to symlink this to `/usr/local/bin/aws` for backwards compatibility. This was slightly less readable than installing from the zip file.

</p>
</details> 

## How to test
- Deploy to CODE ([done](https://riffraff.gutools.co.uk/deployment/view/6b8fe730-9965-4e90-8938-93f484611be6))
- Bake a recipe that uses the `aws-tools` role, it should succeed ([done](https://amigo.code.dev-gutools.co.uk/recipes/arm64-jammy-java11-deploy-infrastructure/bakes/27))

## What is the value of this?
- Using the latest version of the AWS CLI.
- This installation mechanism should allow AMIgo to provide better support for Ubuntu 24.04 as when installing via `pip3` we're greeted with an error introduced in Python 3.12:

```log
This environment is externally managed
╰─> To install Python packages system-wide, try apt install
python3-xyz, where xyz is the package you are trying to
install.

If you wish to install a non-Debian-packaged Python package,
create a virtual environment using python3 -m venv path/to/venv.
Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
sure you have python3-full installed.

If you wish to install a non-Debian packaged Python application,
it may be easiest to use pipx install xyz, which will manage a
virtual environment for you. Make sure you have pipx installed.

See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Version 2 brings some breaking changes - https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration-changes.html#cliv2-migration-changes-breaking. I can't see any immediate issues this would bring, however I could very well be wrong! If the breaking changes do impact us, I'd expect it to manifest in a failed deployment as the user-data script of an EC2 service would fail to run. That is, there wouldn't be any service disruption.


roles/aws-tools/tasks/main.yml
```diff
@@ -3,11 +3,28 @@
...
-  apt: name=python3-pip state=present
+- name: Install unzip
+  apt: name=unzip state=present
 
-- name: Install latest AWS CLI
-  command: pip3 install awscli
+- name: [AWS CLI v2] Create temporary directory
+  file: path=/tmp/awscliv2 state=directory
+
+- name: [AWS CLI v2] Download (aarch64)
```
adamnfish, line 12: Super minor point, but given the only difference between the two "download" steps is the string of the current architecture can this be interpolated into a single step instead of using two conditional ones?

No harm at all, it's fine as-is.

roles/aws-tools/tasks/main.yml
```diff
@@ -3,11 +3,28 @@
...
+  get_url: url=https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip dest=/tmp/awscliv2/awscliv2.zip
+  when: ansible_architecture == "aarch64"
+
+- name: [AWS CLI v2] Download (x86_64)
+  get_url: url=https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip dest=/tmp/awscliv2/awscliv2.zip
+  when: ansible_architecture == "x86_64"
+
+- name: [AWS CLI v2] Extract
+  command: unzip -o /tmp/awscliv2/awscliv2.zip -d /tmp/awscliv2
```
adamnfish, line 21: Why do we need -o, here? If this is a fresh bake in a temporary directory, it should be empty?

Also, can we omit the final `-d` argument since we're unzipping right here?

adamnfish (commented):
The "image bakery" model means we do this work once, properly. With that in mind, should we be verifying the signature as well? The verification mechanism uses pgp, so we'd need gpg installed on the box as well is the only real downside.

The instructions are fairly simple once gpg is installed

* include the AWS public key in this role
* make sure that the public key is imported into the poublic key ring
* also download the signature file for the AWS CLI
* Verify the signature of the downloaded zip file with gpg `gpg --verify awscliv2.sig awscliv2.zip`

I think generally in these image builds if we're downloading stuff off the internet we should be taking the extra step of verifying it, where possible.

Very happy to think about adding this separately though, we don't have those checks now and should focus on unblocking the 24.04 image creation!

roles/aws-tools/tasks/main.yml
akash1810: > Why do we need -o, here? If this is a fresh bake in a temporary directory, it should be empty?

That's what I thought too! However, w/out `-o` the following error happens:

```log
[2024-10-07 19:19:20] arm64-jammy-java11-deploy-infrastructure: fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": ["/tmp/awscliv2/aws/install", "--bin-dir", "/usr/local/bin", "--install-dir", "/usr/local/aws-cli"], "delta": "0:00:02.810203", "end": "2024-10-07 19:19:17.152095", "msg": "non-zero return code", "rc": 1, "start": "2024-10-07 19:19:14.341892", "stderr": "Found preexisting AWS CLI installation: /usr/local/aws-cli/v2/current. Please rerun install script with --update flag.", "stderr_lines": ["Found preexisting AWS CLI installation: /usr/local/aws-cli/v2/current. Please rerun install script with --update flag."], "stdout": "", "stdout_lines": []}
```

I've not found the root cause, but my initial thought was that this role is run multiple times as its [a dependant of others](https://github.com/search?q=repo%3Aguardian%2Famigo%20aws-tools&type=code).

roles/aws-tools/tasks/main.yml
akash1810: I took inspiration from [other roles](https://github.com/search?q=repo%3Aguardian%2Famigo+when%3A+ansible_architecture+%3D%3D+%22x86_64%22&type=code) here.

Personally, I'd find a big ol' bash script easier to reason about (vs multiple steps), but I don't think that's "the Ansible way".

roles/aws-tools/files/aws.pub
akash1810, line 1: The contents of this file are from https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html.

akash1810:
> > With that in mind, should we be verifying the signature as well?
> 
> Turns out this was simpler than anticipated. I've added this now.
> 
> Any advice on how to resolve the following warning welcomed:
> 
> image

Resolved in https://github.com/guardian/amigo/pull/1562/commits/ffbb78f521ec6d82920faa6dc4f943e331af0fe0.

There's still more stdout detail than possibly desired, but we never really read the AMIgo build log so 🤷🏽:

image

roles/aws-tools/tasks/main.yml
```diff
@@ -3,11 +3,84 @@
...
 
-- name: Install latest AWS CLI
-  command: pip3 install awscli
+- name: Install gpg
+  apt: name=gnupg state=present
+
+# It is not uncommon for this role to be run once in a base image, and again in a recipe.
+# There's no point installing the AWS CLI twice, so check if it's already installed.
+- shell: which aws 2>/dev/null || echo aws_cli_not_installed
+  register: aws_cli_installed
```
akash1810, line 15: In this role, we're creating a gpg key for user `AMIgo`. This means on first run, it succeeds but on second it fails as the key we're trying to create already exists.

This `register` syntax was the only way I could see to create a variable in ansible. Ultimately, `aws_cli_installed` is used to detect if this role has already been executed or not.

Joemgru (approved):

akash1810:
This looks to have caused an issue for @guardian/dotcom-platform who were relying on `make` being installed transitively via the AWS CLI v1. They have now made this dependency explicit by directly installing `make`.

PR #1585, opened 2024-12-02
Removes the Ophan Dist S3 bucket dependency from the friendly hostnames role

SHession:
## What does this change?

This PR is intended to remove a direct dependency between the `friendly-hostnames` role and the Ophan Dist S3 bucket, which is currently required to be public for this to work. 

The change takes the resource from the Ophan Dist bucket and provides it directly as a resource in the role. 

[Relates to GChat Thread.](https://chat.google.com/room/AAAAwc4LMKU/Cs6jpX_Iu3s/Cs6jpX_Iu3s?cls=10)

## How to test

I will test this on CODE to confirm it works as expected. 

[Test bake here.](https://amigo.code.dev-gutools.co.uk/recipes/elasticsearch-7-arm-test/bakes/1)
## What is the value of this?

This changes will allow us to remove the S3 dependency of the `friendly-hostnames` role and set the bucket to private. 

Worst case scenario, this could cause issues for our ElasticSearch clusters, namely CAPI and Ophan, the consumers of this role. 

image

### DIFF
--- a/roles/friendly-hostnames/README.md
+++ b/roles/friendly-hostnames/README.md
@@ -7,8 +7,7 @@ and it isn't necessary to name them. Sometimes you want to treat them more like 
 Perhaps they hold data meaning there is a risk associated with indiscriminately killing them and replacing them with 
 another. You may benefit from knowing which server you have up at a given time.
 
-The feature sets the hostname for your server and applies an AWS "Name" tag. The name is chosen from a list of names, 
-defaulting to ~3000 names at `https://s3-eu-west-1.amazonaws.com/ophan-dist/ophan/elasticsearch/hostnames.txt`.
+The feature sets the hostname for your server and applies an AWS "Name" tag. The name is chosen from a list of ~3000 feminists names.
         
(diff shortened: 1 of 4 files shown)


roles/friendly-hostnames/README.md
```diff
@@ -20,13 +19,6 @@ A hostname will be chosen at random and formatted to a
 [valid hostname](https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames), 
 only including the characters [^A-Za-z].  
 
-If you would like to override the default list and supply your own names you will need to specify your hostnames URL, 
-for example
-
-```
-hostnames_url: https://example.com/hostnames.txt
-```
-
```
SHession, line 29: I have checked and this functionality is not being used by any of the consuming recipes.

SHession:
@jonathonherbert, I think it's questionable whether all of these recipes should be using this role if they aren't ES related (which the name suggests): 
image

groakland (approved):
LGTM

PR #1601, opened 2025-01-15
feat(wazuh-agent): Remove the wazuh-agent role from AMIgo

louishather:
## What does this change?

Wazuh is no longer being used to stream ec2 events and package vulnerabilities to InfoSec. Happy to have a chat about this change if anyone has any questions but I think people know it's being wound down. Log sources covered by Wazuh have already moved or are planned to be migrated to InfoSecs new SIEM. 

We'll need to make some changes across other repos, like removing references to wazuh in documentation, and for CDK removing the rules in the EC2 construct allowing traffic to Wazuh. 

The inclusion of the role has been disabled for base images and any recipes where the agent was explicitly included. This was done before the Christmas break, so there shouldn't be many stacks still using AMIs with wazuh-agent included - in any case, there aren't any AMIgo recipes/base images that use this role, so removing it will be fine. 

## What is the value of this?

Role isn't used but some people may include it in new recipes out of habit. Removing the role reduces complexity.

The role was removed from recipes and base images to allow a sunset period, and I checked a few times over that period it hadn't been included in anything new.


louishather:
Pretty sure @philmcmahon and @guardian/investigations don't need this but worth checking before merge.

philmcmahon (approved):

jorgeazevedo (approved):
Sounds good!

jorgeazevedo:
We still open up ports in our ec2 instances for wazuh traffic I think. Is this on your radar @louishather? For example https://github.com/search?q=repo%3Aguardian%2Fcdk%20wazuh&type=code

louishather:
> We still open up ports in our ec2 instances for wazuh traffic I think. Is this on your radar @louishather? For example https://github.com/search?q=repo%3Aguardian%2Fcdk%20wazuh&type=code

Yeah definitely on the radar - along with removing other references to it from docs/various places

jorgeazevedo:
@louishather great 👍

PR #1607, opened 2025-01-22
Add step to whisperx role to cache models required for transcription

philmcmahon:
## What does this change?
In order to run transcription, whisperx relies on a number of different models that need to be fetched from [hugging face](https://huggingface.co) or pytorch. I want to pre-bake all of these models into the AMI so that transcription can run offline.

Fetching the models is a somewhat fiddly process - the best I could do was write this python script to download the models https://github.com/guardian/transcription-service/blob/add-whisperx-support/scripts/download_whisperx_models.py - though I'll hopefully in future add it as a function to whisperx. The script is fetched from the amigo-data buckets, see https://github.com/guardian/transcription-service/pull/130 for details of how it gets there

Also - remove torchvision role which isn't required for transcription.

## How to test
I've tested this on CODE - the resulting AMI was able to transcribe files without an internet connection

### DIFF
--- a/roles/whisperx/tasks/main.yml
+++ b/roles/whisperx/tasks/main.yml
@@ -23,7 +23,7 @@
 #  - https://github.com/guardian/transcription-service/pull/130
 - name: Download models script
   shell: |
-    aws --quiet s3 cp s3://{{ amigo_data_bucket }}/deploy/{{ model_script_stage }}/whisperx-model-fetch/download_whisperx_models.py /tmp/download_whisperx_models.py
+    aws --quiet s3 cp s3://amigo-data-{{ model_script_stage.lower() }}/deploy/{{ model_script_stage }}/whisperx-model-fetch/download_whisperx_models.py /tmp/download_whisperx_models.py
     exit 0
 
 # The script lives here https://github.com/guardian/transcription-service/blob/main/whisperx-model-fetch/download_whisperx_models.py
--- a/roles/whisperx/defaults/main.yml
+++ b/roles/whisperx/defaults/main.yml
@@ -1,3 +1,2 @@
 ---
-amigo_data_bucket: amigo-data-prod
 model_script_stage: PROD


roles/whisperx/tasks/main.yml
```diff
@@ -9,13 +9,22 @@
...
 - name: Install whisperx
   pip:
-    name: whisperx
+    name: "{{ whisperx_package | default('whisperx') }}"
     executable: pip3
 
+- name: Download model fetch script
+  get_url:
+      url: "https://raw.githubusercontent.com/guardian/transcription-service/refs/heads/add-whisperx-support/scripts/download_whisperx_models.py"
+      dest: "/tmp/download_whisperx_models.py"
```
akash1810, line 24: Referencing a file on a feature branch feels a little fragile, as it'll break if/when the branch is deleted. Should `guardian/transcription-service` perform an `aws-s3` deployment with Riff-Raff and AMIgo reference that file?

akash1810: > Do you think that sounds like a good solution?

Yep, this sounds good to me. Might be worth adding inline comments to the file too to explain what's happening for future travellers.

philmcmahon: Good shout, I've made those changes and added the documentation

philmcmahon: (associates transcription pr https://github.com/guardian/transcription-service/pull/130)

roles/whisperx/tasks/main.yml
```diff
@@ -9,13 +9,26 @@
...
+    name: "{{ whisperx_package | default('whisperx') }}"
     executable: pip3
 
+# We use a python script to fetch the models which is owned by the transcription service. See the below PRs for details:
+#  -  https://github.com/guardian/amigo/pull/1607
+#  - https://github.com/guardian/transcription-service/pull/130
+- name: Download models script
+  shell: |
+    aws --quiet s3 cp s3://amigo-data-{{ model_script_stage.lower() }}/deploy/{{ model_script_stage }}/whisperx-model-fetch/download_whisperx_models.py /tmp/download_whisperx_models.py
+    exit 0
```
akash1810, line 27: Is this `exit 0` needed?

roles/whisperx/tasks/main.yml
```diff
@@ -9,13 +9,26 @@
...
-    name: whisperx
+    name: "{{ whisperx_package | default('whisperx') }}"
     executable: pip3
 
+# We use a python script to fetch the models which is owned by the transcription service. See the below PRs for details:
+#  -  https://github.com/guardian/amigo/pull/1607
+#  - https://github.com/guardian/transcription-service/pull/130
+- name: Download models script
+  shell: |
+    aws --quiet s3 cp s3://amigo-data-{{ model_script_stage.lower() }}/deploy/{{ model_script_stage }}/whisperx-model-fetch/download_whisperx_models.py /tmp/download_whisperx_models.py
```
akash1810, line 26: Should we make the bucket a parameter too? Interestingly, the [`cdk-base` role](https://github.com/guardian/amigo/blob/main/roles/cdk-base/defaults/main.yml) uses a bucket w/out the stage suffix.

roles/whisperx/tasks/main.yml
```diff
@@ -9,13 +9,26 @@
...
+    exit 0
+
+# The script lives here https://github.com/guardian/transcription-service/blob/main/whisperx-model-fetch/download_whisperx_models.py
+# If you are changing these parameters it may be helpful to run it locally to test the changes.
+- name: Download whisperx models
+  command: "python3 /tmp/download_whisperx_models.py --whisper-models --diarization-models --torch-align-models --huggingface-token {{ huggingface_token }}"
+  become: yes
```
akash1810, line 33: `become: yes`?!?! Ansible's API is confusing! 😅

akash1810 (approved):
Couple of non-blocking comments.

PR #1617, opened 2025-02-12
Add step to configure postgres apt repository to enable installation of postgresql 16

jonathonherbert:
## What does this change?

`postgresql-16` is not included in the apt repositories available in `[ubuntu-jammy (22.04 LTS)](https://amigo.gutools.co.uk/base-images/ubuntu-jammy%20(22.04%20LTS))`. This PR updates apt with the postgres repository as an initial step, to ensure `postgresql-16` is available.

## How to test

Run the bake locally with `multipass` per the README. It should succeed. Tested locally:

Screenshot 2025-02-12 at 12 20 45

## What is the value of this?

postgres 16 available for Kong.

The image will be well tested before being deployed to PROD services. A failed or incorrectly configured bake is low risk.

### DIFF
--- a/roles/kong/tasks/main.yml
+++ b/roles/kong/tasks/main.yml
@@ -5,6 +5,7 @@
       - postgresql-common
 
 - name: Configure postgres repository to enable download of latest postgres client
+  # Per https://wiki.postgresql.org/wiki/Apt
   script: /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y
 
 - name: Install other Kong dependencies
--- a/roles/kong/tasks/main.yml
+++ b/roles/kong/tasks/main.yml
@@ -1,5 +1,13 @@
 ---
-- name: Install Kong's dependencies
+- name: Install initial postgres dependencies
+  apt:
+    name:
+      - postgresql-common
+
+- name: Configure postgres repository to enable download of latest postgres client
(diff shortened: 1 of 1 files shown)


roles/kong/tasks/main.yml
```diff
@@ -1,5 +1,13 @@
 ---
-- name: Install Kong's dependencies
+- name: Install initial postgres dependencies
+  apt:
+    name:
+      - postgresql-common
+
+- name: Configure postgres repository to enable download of latest postgres client
```
JustinPinner, line 7: Is there a risk that this could download a version that's incompatible with the specific `postgresql-client-{{postgres_version}}` [here](https://github.com/guardian/amigo/pull/1617/files#diff-36a56733dc57a3473a271b4bf2933be4194cd6aed093b0324626ecda7f9d3973R18)?

jonathonherbert: Absolutely, there's no necessary correlation between the two — as before with e.g. apt no longer including `postgresql-client-16`, there are implicit dependencies that I'm not sure how to avoid here.

jonathonherbert: (I've added a comment linking the postgres apt guidance in ebec37eab572ce32d8e1e1c63b4b588bc2c041c1, so if this changes, there's a breadcrumb)

JustinPinner (approved):
👍

PR #1620, opened 2025-02-12
Correct Kong URL with additional parameter for Ubuntu release

jonathonherbert:
## What does this change?

Corrects the URL from which we fetch Kong, with an additional parameter for the Ubuntu release — we're now on Jammy.

Also adds the missing configuration file for 3.9.0.

## How to test

The bake should succeed with the appropriate parameters bumped for 3.9.0:

```
---
kong_version: 3.9.0
kong_user: content-api
postgres_version: 16
kong_arch: arm64
kong_ubuntu_release: jammy
kong_repo: gateway-39
```

Tested locally 👍 

Screenshot 2025-02-12 at 17 32 24

### DIFF
--- a/roles/kong/defaults/main.yml
+++ b/roles/kong/defaults/main.yml
@@ -3,4 +3,5 @@ kong_version: 3.1.1
 kong_user: content-api
 postgres_version: 12
 kong_arch: arm64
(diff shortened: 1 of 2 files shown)


simonbyford (approved):
LGTM

PR #1622, opened 2025-02-13
Add conf file for kong 3.9.0

jonathonherbert:
## What does this change?

Add conf file for kong 3.9.0.

## How to test

Bake should succeed.


JustinPinner (approved):
👍 Will assume the defaults are all appropriate - I haven't diffed the previous config with the new one to check for new settings etc, but as it's a dual stack setup, we can tease out any such differences as and when they show up.

PR #1627, opened 2025-03-03
Expose less information to kong clients

jonathonherbert:
## What does this change?

Kong 3.9 exposes more information than we'd like to its clients. This PR adds configuration to:

- Remove headers that expose version and latency information, as well as request id (Kong <3.9 also exposed version and latency information)
- Remove request id from error templates

## How to test

- Deploy an AMI to CODE. Requests should no longer include the above information.

## What is the value of this?

Reduced information for potential attackers.

- The format of the default error response is changing. We don't that's important for well-behaved clients.


JustinPinner (approved):
Watched this take shape - am happy to approve 👍

PR #1630, opened 2025-03-13
[await-tags] Provide authentication to instance metadata API

davidfurey:
## What does this change?

Calls to the instance metadata API without authentication are no longer accepted. Adding the token effectively updates this from using IMDSv1 to IMDSv2. This should resolve FSBP EC2.8 for a number of services in the editorial tools space.

NB- this role is still deprecated, but it is quicker to upgrade it for FSBP compliance than to convince ourselves that it isn't needed

## How to test

- Bake the `ditorial-tools-focal-java11-ARM-WITH-cdk-base` recipe
- Redeploy workflow archiver
- Check logs and see that `AWS Tags not found after 1 minute` does not appear but `AWS tags loaded` does
- Check https://metrics.gutools.co.uk/d/de156avqnutj4d/am-i-using-imdsv1-or-imdsv2?orgId=1&refresh=5s&var-AwsAccount=lpcN6XMMz&var-AwsAccountId=753338109777&var-Stack=workflow&var-Stage=PROD&var-App=archiver&var-ASG=workflow-PROD-ProleAutoscalingGroup-EHT3JYOHS8NF&var-RiffRaffProject=&from=now-30d&to=now to see that no calls are made to IMDSv1

## What is the value of this?

Resolves FSBP EC2.8

This script is already broken because the existing INSTANCE_ID call fails, so this PR cannot make things worse.

### DIFF
--- a/roles/aws-await-tags/templates/await_tags.sh
+++ b/roles/aws-await-tags/templates/await_tags.sh
@@ -1,7 +1,8 @@
 #!/bin/bash -e
 echo "Awaiting AWS tags..."
 
-INSTANCE_ID=$(curl --silent http://169.254.169.254/latest/meta-data/instance-id)
+TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+INSTANCE_ID=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" --silent http://169.254.169.254/latest/meta-data/instance-id)
 
 for i in {1..12}
 do


jorgeazevedo (approved):
Nice one, I think this should work! I ran the curl calls in a real instance and they behaved as expected

davidfurey:
Thanks for the review!

emdash-ie:
Have started [bake 114 of editorial-tools-focal-java11-ARM-WITH-cdk-base](https://amigo.gutools.co.uk/recipes/editorial-tools-focal-java11-ARM-WITH-cdk-base/bakes/114) to test that this works as we expect.

emdash-ie:
Bake finished, [deploying archiver to CODE now](https://riffraff.gutools.co.uk/deployment/view/fc97b970-f49e-44d2-b038-29914858e719).

emdash-ie:
The change doesn’t appear to have fixed the tag lookup, it seems: [the logs still show the message “AWS tags not found after 1 minute”](https://logs.gutools.co.uk/s/editorial-tools/app/r/s/WzeGI). Will check for usages of the IMDSv1 API though, as that should still have come down even if the script isn’t working.

emdash-ie:
It looks like [archiver in CODE](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#AutoScalingGroupDetails:id=workflow-CODE-ArchiverAutoscalingGroup-2VVA9RSNO89O;view=instanceManagement) is still using the old AMI (bake 113). I think my deploy was insufficient, possibly I needed to deploy the cloudformation?

emdash-ie:
Apparently it can take time for AMIs to be visible in the right account, but a normal deploy should be fine once it is visible. I have confirmed that [the right AMI](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#ImageDetails:imageId=ami-034306c8a3eb65c31) is visible, and started [a new deploy](https://riffraff.gutools.co.uk/deployment/view/fe882d6e-c72c-4221-a3a6-4541f70e6af7). Hopefully this one should do it!

emdash-ie:
The logs look good this time: https://logs.gutools.co.uk/s/editorial-tools/app/r/s/49bKq

emdash-ie:
[The metric also looks good](https://metrics.gutools.co.uk/d/de156avqnutj4d/am-i-using-imdsv1-or-imdsv2?orgId=1&var-AwsAccount=lpcN6XMMz&var-AwsAccountId=753338109777&var-Stack=workflow&var-Stage=CODE&var-App=archiver&var-ASG=workflow-CODE-ArchiverAutoscalingGroup-2VVA9RSNO89O&var-RiffRaffProject=&from=1741884000000&to=1741885259000): there’s a datapoint for MetadataNoTokenRejected at 16:40 which is from the previous deploy, but none after that.

PR #1634, opened 2025-03-25
The recipe `hasUsage` function now includes launch template checks

adamnfish:
## What does this change?

Launch template support was added last year, but it doesn't look like the UI was fully updated to reflect this. This means a manual tidy up job might accidentally remove recipes that appear to be unused despite being referred to by launch templates.

We add this check to the logic, so that the "Unused recipes" section of the recipes index is correctly populated. This also improves the usage checks on base images page.

## How to test

This adds a couple more tests for the `hasUsage` function, to cover non-instance usages explicitly.

## What is the value of this?

This should help prevent manual clear-ups from removing recipes that are still in use.

> **Note**


kelvin-chappell (approved):
Well spotted! 👍

PR #1801, opened 2026-03-13
Remove unused cloudformation parameter from image copier stack

philmcmahon:
## What does this change?
Somewhere along the road of the CDK migration (I'm assuming) we stopped respecting the KmsKeyArn cloudformation parameter in the image copier CDK. I initially set about to get it working again, but, because I was scared of pushing out a change that would impact teams who had unknowingly set that parameter, I've instead removed it entirely.

## How to test
This should be a noop, I've tested it on the investigations account

### DIFF
--- a/cdk/lib/__snapshots__/image-copier-lambda.test.ts.snap
+++ b/cdk/lib/__snapshots__/image-copier-lambda.test.ts.snap
@@ -15,11 +15,6 @@ exports[`the lambda stack matches the snapshot 1`] = `
       "Description": "The SNS topic to subscribe to",
       "Type": "String",
     },
-    "KmsKeyArn": {
-      "Default": "",
-      "Description": "Override the default KMS key if required",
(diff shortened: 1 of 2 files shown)


jorgeazevedo (approved):

jorgeazevedo:
Do you know how this is deployed? Is it stacksets via the root account?

philmcmahon:
> Do you know how this is deployed? Is it stacksets via the root account?

yes it is, I don't mind doing the deploy as I currently still have root account access. Though it is a no-op so maybe we don't need to bother?

jorgeazevedo:
I just realised now that I merged an update to Java 21 but forgot to deploy it 😬https://github.com/guardian/amigo/commit/5e645b682594b8c87d36769e45cb02ad276f042d

I should take that one on in case of any issues. Feel free to merge this PR, then I'll just deploy from main and include your change as well

PR #1856, opened 2026-06-17
Implement switch to disable scheduled bakes on PROD if problems occur

JuliaBrigitte:
## What does this change?

Implement and document a configuration switch in amigo to disable scheduled bakes. 
Manually created the parameter for CODE and PROD set to false and true.

## How to test

Deploy to CODE and set temporarily to true.

## What is the value of this?

The use case is when we detect an issue with a component that is being spread to every new bake.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1215541806167213


app/components/AppComponents.scala
```diff
@@ -243,7 +243,15 @@ class AppComponents(context: Context, identity: AppIdentity)
...
-    val enabled = stage == "PROD" // don't run scheduled bakes on dev machines
+    // allow scheduled bakes to be disabled via config
+    val enabled = configuration
+      .get[Option[String]]("amigo.scheduledBakes.enabled")
+      .flatMap(_.trim.toLowerCase match {
+        case "true"  => Some(true)
+        case "false" => Some(false)
+        case _       => None
+      })
+      .getOrElse(false)
```
akash1810: minor: Could this be simplified:

```scala
val enabled: Boolean = configuration
  .getOptional[String]("amigo.scheduledBakes.enabled")
  .getOrElse("false")
  .equalsIgnoreCase("true")
```

README.md
```diff
@@ -35,6 +35,11 @@ When running in an environment that is not `PROD` there is an option to `Bake wi
 This passes the `-debug` flag through to packer which saves a copy of the SSH key in AMIgos working directory. This makes 
 it possible to SSH onto the instance that is being used to build the AMI. 
 
+## Troubleshooting problems with scheduled bakes
+
+If we detect an issue in a component that would affect all new bakes, we can temporarily disable scheduled bakes.
+Just set the Parameter Store switch `amigo.scheduledBakes.enabled` to false.
+
```
akash1810, line 45: Could we provide a script for this? I imagine if we're reading these docs, we're triaging an incident and would want to react quickly. A script to copy/paste would be faster than navigating the AWS console. WDYT?

akash1810: Would AMIgo also need to be redeployed to pick up this configuration change?

JuliaBrigitte: I think that must be the case, it would be really good to check for configuration changes at runtime to be able to just change the parameter store value and the scheduled bakes are stopped. Thanks @akash1810 I will look at it.

app/schedule/ScheduledBakeRunner.scala
```diff
@@ -8,7 +8,7 @@ import services.{AmiMetadataLookup, Loggable, PrismData}
 
 class ScheduledBakeRunner(
     stage: String,
-    enabled: Boolean,
+    scheduledBakesEnabled: () => Boolean,
```
akash1810: Having this as a function means each scheduled bake makes an AWS API call. I can't work out if that's expensive or not 🤔.

akash1810 (approved):

akash1810 (dismissed):
I've pushed a change to the bash script to refactor for brevity.

akash1810:
[I've deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/431ee1cb-6734-48f7-8f5a-3c218c76f3c7). I've set a recipe in CODE to bake with a schedule of `0 * * ? * TUE *` (every minute, every hour, every Tuesday). The logs suggest the configuration is not being read at runtime, but only once at app start. I've therefore pushed a change to revert some of the changes here and to add a note that a deployment is needed for the changes to take effect.

image

akash1810 (approved):
I've [deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/3e11f1c1-e283-40a1-9fd0-9212547cc18f), adjusted the configuration and [deployed to CODE again](https://riffraff.gutools.co.uk/deployment/view/7885f537-5b9e-466c-8afa-7a6689a33f54). The logs show expected behaviour:

Image

PR #1896, opened 2026-09-16
chore: Add dev container config

kelvin-chappell:
## What does this change?
This makes some config changes so that the Play app can easily be run in a dev container.

1. Tool versions file contains all the tools required to work on this codebase
2. The [devenv](https://github.com/guardian/devenv) generated files have been added
3. Removed redundant `nvm` config
4. Fetching dev config from SSM parameters (same as other stages).
5. A little code change to avoid having multiple credentials provider chains.

## How to test
Run locally in a dev container.
- [ ] Tested on Code


.devcontainer/devenv.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - mise
+  - scala
+  - node
+  - github-copilot
+
+forwardPorts:
+  - 9000
+
+mounts:
+  - "source=${localEnv:HOME}/.gu/amigo,target=/home/vscode/.gu/amigo,readonly,type=bind,consistency=cached"
```
adamnfish-gu, line 17: Is this mount needed? If we're going the "download in a setup script" route then maybe we can skip this step? Particularly if we've all lost the config file from our Macs after upgrading laptops anyway!

.tool-versions
```diff
@@ -1 +1,5 @@
 java corretto-21
+sbt 1
+node 24
```
adamnfish-gu, line 3: Flagging that this also updates node, but I'm sure that'll be fine here.

conf/application.conf
```diff
@@ -10,9 +10,6 @@ play.application.name = "amigo"
 play.filters.csp.directives.script-src = "'self' https://maxcdn.bootstrapcdn.com/bootstrap/ https://code.jquery.com/ https://cdnjs.cloudflare.com/ajax/libs/showdown/1.4.3/ 'sha256-n5iMc+TYRqvKiJKPzco5gfPgoRaBlwBZqb+yNH4JTLo=' 'sha256-/PoztUMLJgk2dbb1A1+S/VWnwljIBNIey/XF0k0ln3g=' 'sha256-I7IpBXTYPPnBzwsqYsQEQFePIjnD0r/pF+9dmU/MQkI='"
 
 google {
-  # These are dev credentials, only valid when running on localhost
```
adamnfish-gu, line 13: I guess this is provided from the local config file instead, now?

Might be nice to add this to the build.sbt so that people don't need to manually include the `-D` flag when they invoke sbt, like we do elsewhere? Fine to do multiple rounds of iterating towards a working dev environment though!

adamnfish-gu (commented):
I'm not sure this will work, because we're currently still expecting to use Brew to install some of the OS runtime dependencies?

The mount setup is interesting, because I think that's something we've decided to avoid? Maybe it's worth doing this via S3 straight away, so that we don't create devcontainers with mounts while we continue to work on the dev environment?

No harm to split it up into small chunks, and maybe this is just the first step 👍

.tool-versions
kelvin-chappell: Good point!  I think so.

.devcontainer/devenv.yaml
kelvin-chappell: The amigo mount is because I glossed over that there's one last piece of config locally, which is the service account key.  I couldn't move this without making more code changes or changes to other environments.  But I don't provide any means of fetching it.

conf/application.conf
kelvin-chappell: This is now in SSM params.

jorgeazevedo (approved):

.devcontainer/devenv.yaml
kelvin-chappell: Having discussed with @jorgeazevedo, he will add a dev container setup script to do this and remove the mount.

PR #1898, opened 2026-09-21
Add system-wide tool installation to uv role

philmcmahon:
## What does this change?
Extends the existing `uv` role with a `uv_tools` list so we can install uv tools - see https://docs.astral.sh/uv/concepts/tools/ for details. Each tool accepts a package requirement, an optional Python version, and optional installation arguments. This makes it possible to install tools that have differing python version requirements.

In the transcription service we'll be using this to install whisperx (python 3.10) and ocrmypdf (python 3.12+) alongside one another

## Change to uv installation
The only change here that will affect existing uses of this role is the move of `uv` to be installed in its own `venv` rather than being installed into the os python environment. This was necessary to avoid this kind of error when running on ubuntu 26.04:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
python3-xyz, where xyz is the package you are trying to
install.

If you wish to install a non-Debian-packaged Python package,
create a virtual environment using python3 -m venv path/to/venv.
Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
sure you have python3-full installed.

If you wish to install a non-Debian packaged Python application,
it may be easiest to use pipx install xyz, which will manage a
virtual environment for you. Make sure you have pipx installed.

See /usr/share/doc/python3.14/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

## How to test
Tested here https://amigo.code.dev-gutools.co.uk/recipes/cuda-ami-test on amigo CODE - I verified that I could transcribe something using whisperx on the output AMI.

The readme for the role includes detailed examples of using the role.


philmcmahon:
Update - I checked in with @joelochlann  (who's team is the only user of this role) and he's happy for this to go out

jorgeazevedo (approved):

PR #1899, opened 2026-09-22
Add script to fetch necessary certificate

jorgeazevedo:
## What does this change?

Adds a setup script to download the certificate. This obviates need for mount the home folder.


kelvin-chappell (approved):
There's some coupling between code and dev stages that we could break out in future PRs but this looks good 👍

